PR #137, opened 2023-06-15
Change laion metric to L2

XuanYang-cn:
The groudthruth is caculated with L2, change to L2 will skip normalization for DBs don't support COSIN metric.


zhuwenxing (approved):
/lgtm

PR #168, opened 2023-06-28
Update elastic_cloud runner to adhere to best practices

benwtrent:
Updates elastic_cloud runner to adhere to best practices

 - disabling vectors in _source
 - force merging down to a single segment

I verified elastic_cloud works when testing locally.

The rest of the changes in the commit were because I simply ran automatic formatting fixing before committing, I can revert those if required :)

### DIFF
--- a/vectordb_bench/backend/clients/elastic_cloud/elastic_cloud.py
+++ b/vectordb_bench/backend/clients/elastic_cloud/elastic_cloud.py
@@ -1,4 +1,5 @@
 import logging
+import time
 from contextlib import contextmanager
 from typing import Iterable, Type
 from ..api import VectorDB, DBCaseConfig, DBConfig, IndexType
@@ -56,7 +57,7 @@ class ElasticCloud(VectorDB):
     def init(self) -> None:
         """connect to elasticsearch"""
         from elasticsearch import Elasticsearch
-        self.client = Elasticsearch(**self.db_config, request_timeout=30)
+        self.client = Elasticsearch(**self.db_config, request_timeout=180)
 
         yield
         # self.client.transport.close()
@@ -65,8 +66,9 @@ class ElasticCloud(VectorDB):
 
     def _create_indice(self, client) -> None:
         mappings = {
+            "_source": {"excludes": [self.vector_col_name]},
             "properties": {
-                self.id_col_name: {"type": "integer"},
+                self.id_col_name: {"type": "integer", "store": True},
                 self.vector_col_name: {
                     "dims": self.dim,
                     **self.case_config.index_param(),
@@ -137,8 +139,16 @@ class ElasticCloud(VectorDB):
         }
         size = k
         try:
-            search_res = self.client.search(index=self.indice, knn=knn, size=size)
-            res = [d["_source"][self.id_col_name] for d in search_res["hits"]["hits"]]
+            res = self.client.search(
+                index=self.indice,
+                knn=knn,
+                size=size,
+                _source=False,
+                docvalue_fields=[self.id_col_name],
+                stored_fields="_none_",
+                filter_path=[f"hits.hits.fields.{self.id_col_name}"],
+            )
+            res = [h["fields"][self.id_col_name][0] for h in res["hits"]["hits"]]
 
             return res
         except Exception as e:
@@ -147,7 +157,16 @@ class ElasticCloud(VectorDB):
 
     def optimize(self):
         """optimize will be called between insertion and search in performance cases."""
-        pass
+        assert self.client is not None, "should self.init() first"
+        self.client.indices.refresh(index=self.indice)
+        force_merge_task_id = self.client.indices.forcemerge(index=self.indice, max_num_segments=1, wait_for_completion=False)['task']
+        log.info(f"Elasticsearch force merge task id: {force_merge_task_id}")
+        SECONDS_WAITING_FOR_FORCE_MERGE_API_CALL_SEC = 30
+        while True:
+            time.sleep(SECONDS_WAITING_FOR_FORCE_MERGE_API_CALL_SEC)
+            task_status = self.client.tasks.get(task_id=force_merge_task_id)
+            if task_status['completed']:
+                return
 
     def ready_to_load(self):
         """ready_to_load will be called before load in load cases."""


vectordb_bench/backend/clients/elastic_cloud/elastic_cloud.py
```diff
@@ -145,7 +147,16 @@ def search_embedding(
 
     def ready_to_search(self):
         """ready_to_search will be called between insertion and search in performance cases."""
-        pass
+        assert self.client is not None, "should self.init() first"
+        self.client.indices.refresh(index=self.indice)
+        force_merge_task_id = self.client.indices.forcemerge(index=self.indice, max_num_segments=1, wait_for_completion=False)['task']
```
liliu-z, line 152: Does it mean to merge all vector data into a single segment?

liliu-z, line 152: Looks like we turn the `wait_for_completion` into False. Just curios why we don't make it a sync call by turning into True?

vectordb_bench/backend/clients/elastic_cloud/elastic_cloud.py
```diff
@@ -145,7 +147,16 @@ def search_embedding(
...
+        log.info(f"Elasticsearch force merge task id: {force_merge_task_id}")
+        SECONDS_WAITING_FOR_FORCE_MERGE_API_CALL_SEC = 30
+        while True:
+            time.sleep(SECONDS_WAITING_FOR_FORCE_MERGE_API_CALL_SEC)
+            task_status = self.client.tasks.get(task_id=force_merge_task_id)
```
liliu-z, line 157: Any estimation how long it will take, since we are considering to introduce some timeout mechanism for all tasks.

liliu-z:
@benwtrent Thanks a lot for your contribution!!

Do you have any test result number we can direct check for this optimization? It is fine if we don't have any, we can rebenchmark later.

vectordb_bench/backend/clients/elastic_cloud/elastic_cloud.py
benwtrent: @liliu-z correct. All the benchmarks I can see here are "index everything" then "how do we do on searching after the index is built". This is the best practice for such a case.

benwtrent: > Just curios why we don't make it a sync call by turning into True?

This can take many minutes. Its not worth keeping the connection open.

vectordb_bench/backend/clients/elastic_cloud/elastic_cloud.py
benwtrent: Many minutes

liliu-z:
@benwtrent Thanks for your explanation. We will manually verify the correctness and performance, and then I will check this in. Sorry that we don't have any auto-checking mechanism for now.

liliu-z:
@benwtrent Can you plz double check, On a 2.5c ES Cloud it takes 10 hrs instead 8 after this change. And 5 hrs of them are used to do optimization.

benwtrent:
@liliu-z `2.5c ES Cloud` I am unfamiliar with that deployment type What is the RAM and vCPU? I tested with `GCP.ES.DATAHOT.N2.68X16X45`

alwayslove2013:
@benwtrent up to 2.5 vCPU and 8g RAM

alwayslove2013:
Hello @benwtrent~
We tested your latest strategy.
- Machine: We didn't find the same test machine as yours (16c64g) on ElasticCloud, we finally chose 10c64g and 20c128g (Noted that 20c128g displayed as  2 * 10c64g in the system).
- Dataset: 1M data * 768dim.

they got good performance, but we also noticed that they both took a lot of time to load data.
- 10c64g, insert took 2.5 hours, compact took 4 hours.
- 20c128g (10c64g * 2), insert took 3.5 hours, compact took 3.5 hours.

We try to introduce the timeout mechanism to make the benchmark results closer to the actual usage scenarios, and obviously the optimize time should not be too long. Sorry we are not familiar with ES, could you help us to make it faster?
https://github.com/zilliztech/VectorDBBench/blob/16d60d58a5972ce351f9f3d8882f64bf53f42d7e/vectordb_bench/__init__.py#L22-L29

benwtrent:
@alwayslove2013 thanks for continuing to iterate on this!

> We try to introduce the timeout mechanism to make the benchmark results closer to the actual usage scenarios, and obviously the optimize time should not be too long. Sorry we are not familiar with ES, could you help us to make it faster?

I used the `optimize` functionality to build a single graph. So, for the fastest kNN searches with Elasticsearch:

 - Index the data, this will result in multiple HNSW graphs over multiple Lucene segments
 - Force-merge to a single segment, this then has Lucene build a single segment.

The force-merge time is indeed painful and we are working on making that better. Additionally, we are working towards improving performance performance with multiple-segments (thus obviating the force-merge).

As for:

> We try to introduce the timeout mechanism to make the benchmark results closer to the actual usage scenarios,

It seems to me that an actual usage scenario would be parallel indexing while querying. Right now the benchmarks assume indexing data and then searching it. "Real usage" seems like indexing data is continuous, and users query previously indexed data from the same indices. In this scenario, force-merging doesn't make sense. 

But in the "I am loading a ton of data that isn't changing", taking a while to optimize for a continually faster query load make sense to me. But, I agree it shouldn't take as long as it does currently.

I think measuring querying from the same index while indexing new data is a better than arbitrary timeouts. We have similar benchmarks within our own benchmarking via Rally: https://elasticsearch-benchmarks.elastic.co/#tracks/dense_vector/nightly/default/90d (see: `nightly-dense_vector-add-4g-1node-knn-search-100-1000-concurrent-with-indexing-latency`, I think the task name long enough 🤣)

Benchmarking is a difficult endeavor, and benchmarking multiple different platforms while trying to ensure fairness is even more difficult. Thank you for your work!

liliu-z:
> @alwayslove2013 thanks for continuing to iterate on this!
> 
> > We try to introduce the timeout mechanism to make the benchmark results closer to the actual usage scenarios, and obviously the optimize time should not be too long. Sorry we are not familiar with ES, could you help us to make it faster?
> 
> I used the `optimize` functionality to build a single graph. So, for the fastest kNN searches with Elasticsearch:
> 
> * Index the data, this will result in multiple HNSW graphs over multiple Lucene segments
> * Force-merge to a single segment, this then has Lucene build a single segment.
> 
> The force-merge time is indeed painful and we are working on making that better. Additionally, we are working towards improving performance performance with multiple-segments (thus obviating the force-merge).
> 
> As for:
> 
> > We try to introduce the timeout mechanism to make the benchmark results closer to the actual usage scenarios,
> 
> It seems to me that an actual usage scenario would be parallel indexing while querying. Right now the benchmarks assume indexing data and then searching it. "Real usage" seems like indexing data is continuous, and users query previously indexed data from the same indices. In this scenario, force-merging doesn't make sense.
> 
> But in the "I am loading a ton of data that isn't changing", taking a while to optimize for a continually faster query load make sense to me. But, I agree it shouldn't take as long as it does currently.
> 
> I think measuring querying from the same index while indexing new data is a better than arbitrary timeouts. We have similar benchmarks within our own benchmarking via Rally: https://elasticsearch-benchmarks.elastic.co/#tracks/dense_vector/nightly/default/90d (see: `nightly-dense_vector-add-4g-1node-knn-search-100-1000-concurrent-with-indexing-latency`, I think the task name long enough 🤣)
> 
> Benchmarking is a difficult endeavor, and benchmarking multiple different platforms while trying to ensure fairness is even more difficult. Thank you for your work!

Hello @benwtrent,

Thank you for your insightful feedback. I completely agree that while our current batch processing cases (querying post-indexing) can mimic certain real-world scenarios, we indeed need to incorporate more complex and comprehensive cases for better representation of real-world use cases. This includes, as you've pointed out, streaming processing cases (querying concurrent with indexing). In fact, we initially planned for this, and I've created a new issue on our GitHub to track this: https://github.com/zilliztech/VectorDBBench/issues/182.

As for batch processing, while optimization is welcome, an overly prolonged optimization period may not fit the majority of real-world scenarios. Over-optimization can result in other challenges that impact real production systems but might remain undetected by benchmarks. For instance, forcing a massive data merge into a single segment can cause issues related to failure recovery and data migration within a distributed system. Therefore, we're working to filter out all strategies that won't apply to most production cases. I also agree with you that instead of merely relying on timeout, we should strive for more nuanced ways to identify cases that aren't ready for production. Rest assured, we're doing our best in this direction.

Your suggestions and contributions are invaluable in enhancing this benchmark to serve all VectorDB users better!

By the way, if you believe this approach is a best practice for ES usage, please let us know. We can merge it and put the benchmark results in our standard results with indicating the build/optimization time. I believe this would also ensure fairness and assist the community.

benwtrent:
Hey @liliu-z && @alwayslove2013 ! I updated the client. It still force-merges to one segment. But as we make improvements we will update this benchmark for sure. The community is missing a macro-benchmark (ann-benchmarks is a really good micro-benchmarking suite for algorithms). 

This one might fit the bill over time.

alwayslove2013 (approved):

alwayslove2013:
Hello @benwtrent ~
We briefly tested your latest PR and it performed pretty well. Next, we'll add it to the benchmark, run a full test, and update the performance. But before merging it, could you please remove the formatting part first, it seems like there is some conflict.

alwayslove2013:
Hello @benwtrent~
We test the 10M*768dim dataset on 10c64g ElaticCloud Service. The insert time is much shorter than before but the optimization time is too long, even more than the insert time. According to our time limitation, it will be labeled "timeout" and its performance will not be displayed.
https://github.com/zilliztech/VectorDBBench/blob/675b48f0d2b5561886af17dd897d789bd5e76f5b/vectordb_bench/__init__.py#L22-L29

Could you please replace your optimization strategy? I also found that after inserting data just sleeping for a while also get a much better performance. Hope it can help you.

benwtrent:
@alwayslove2013 honestly, the elasticsearch optimization occurs in the background in a separate threadpool. In the typical use case, users will still search their data while its occurring. It does not block users from searching or indexing additional data. It is not mandatory for search. I am not sure how to communicate that within the benchmark. But, since this is a benchmark to measure "best speeds", it only seems reasonable to allow it to finish before measuring search latency.

Again, I don't know how to indicate this in a benchmark. While optimizations can take a while, it doesn't block any real work (even searches).

I am curious, how did y'all determine the "appropriate timeout" values here?

I will remove the changes to the other files :). I ran the auto-formatter and I guess other changes have occurred since then.

liliu-z:
> @alwayslove2013 honestly, the elasticsearch optimization occurs in the background in a separate threadpool. In the typical use case, users will still search their data while its occurring. It does not block users from searching or indexing additional data. It is not mandatory for search. I am not sure how to communicate that within the benchmark. But, since this is a benchmark to measure "best speeds", it only seems reasonable to allow it to finish before measuring search latency.
> 
> Again, I don't know how to indicate this in a benchmark. While optimizations can take a while, it doesn't block any real work (even searches).
> 
> I am curious, how did y'all determine the "appropriate timeout" values here?
> 
> I will remove the changes to the other files :). I ran the auto-formatter and I guess other changes have occurred since then.

Hi @benwtrent 
Thank you for your contributions. Would you mind squashing these commits into one for better readability? 

>But, since this is a benchmark to measure "best speeds", it only seems reasonable to allow it to finish before measuring search latency.

Regarding the timeout consideration, allow me to clarify. Our benchmark isn't simply about attaining the fastest speeds; it's about achieving optimal speeds within specific limits. These restrictions are designed to simulate real-world scenarios. For example, if one system need to take 10x longer than most of others to finish the benchmark, we need to let user know this and find a way to mark uncompetitive in the scoring system.

IMHO, there are two ways to ensure that a benchmark case reflects a real-world scenario:
1. Set limits.
2. Incorporate all aspects into a comprehensive scoring system.
For our case, we opted for the first solution concerning data preparation time because it's the most straightforward to implement. Also, many other benchmarks follow similar rules. For example, ann-benchmark doesn't allow the building of an index to take too long, this is the reason why the `NGT-qg` index does not have any result in the `GIST` test.

> I am curious, how did y'all determine the "appropriate timeout" values here?

You asked how we determined the "appropriate timeout" values. We tried to avoid timing out for most of the systems we tested, but we also wanted to ensure the process wasn't too slow (within ~5-7x of the median). For example, in our case, we set a time limit for the entire preparation procedure (build + optimization):
**1M data*: Out of **20** systems, **15** take less than **3k** seconds, **19** take less than **10k** seconds, and only ES 10 CPU takes **~25k** seconds. The median is **~1.7k** seconds. Hence, we set the limit to **~10k** seconds to accommodate all systems except ES.
**10M data**: Out of **8** systems, **6** take less than **15k** seconds, **7** take less than **20k** seconds, and only ES 10 CPU takes **~250k** seconds. The median is **~15k** seconds. Therefore, we set the limit to **~100k** seconds.

There are two potential solutions:
1. We could use a larger ES host to see if optimization can be expedited.
2. We could try to control the level of optimization to fit within the timeout.

----

In any case, let's commit this code first. Here are two improvements we could consider to minimize confusion:

1. Consider how to score systems based on both performance and building time. This would allow us to freely adjust the build time and find the right balance.
2. Allow users to disable the timeout check in the config for a more flexible benchmarking experience.

Thanks!

liliu-z:
> 

Let us take a quick test and check it in. Appreciate it for your contribution!

PR #178, opened 2023-07-05
Renew standard results of zillizcloud

XuanYang-cn:
We retested all metrics of zillizcloud of
zc production instance.

The QPS increased and the latency is more reproducible.


XuanYang-cn:
image
image
image
image

alwayslove2013 (approved):

PR #201, opened 2023-07-21
Chroma client

s-gruneberg:
I made a chroma client in case you are interested. All unit tests passing. I am running the database in docker client/server but it can be configured to change to running in process easily. There's some configuration notes in the test_chroma.py file.

### DIFF
--- a/vectordb_bench/backend/clients/chroma/config.py
+++ b/vectordb_bench/backend/clients/chroma/config.py
@@ -1,0 +1,14 @@
+from pydantic import SecretStr
+from ..api import DBConfig
+
+class ChromaConfig(DBConfig):
+    password: SecretStr
+    host: SecretStr
+    port: int 
+
+    def to_dict(self) -> dict:
+        return {
+            "host": self.host.get_secret_value(),
+            "port": self.port,
+            "password": self.password.get_secret_value(),
+        }
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -39,7 +39,8 @@ dependencies = [
     "psutil",
     "polars",
     "pgvector",
-    "sqlalchemy"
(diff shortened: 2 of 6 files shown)


vectordb_bench/backend/clients/chroma/config.py
```diff
@@ -0,0 +1,16 @@
+from pydantic import SecretStr
+from ..api import DBConfig
+
+class ChromaConfig(DBConfig):
+    name: str = ""
+    password: str = ""
+    host: str = ""
```
XuanYang-cn, line 7: Please use Pydantic SecretStr Type, DB config will be recorded in results, we don't want any sensitive infomations exposed to everyone.
ref:  https://github.com/zilliztech/VectorDBBench/blob/675b48f0d2b5561886af17dd897d789bd5e76f5b/vectordb_bench/backend/clients/milvus/config.py#L6

vectordb_bench/backend/clients/chroma/config.py
```diff
@@ -0,0 +1,16 @@
+from pydantic import SecretStr
+from ..api import DBConfig
+
+class ChromaConfig(DBConfig):
+    name: str = ""
```
XuanYang-cn, line 5: I don't notice any usage of the param `name`, please remove it if it's not essential for connecting to the server.

XuanYang-cn (changes requested):

XuanYang-cn:
It would be better if you could squish all of these changes together into one single commit.

This will make it easier to track changes and maintain a more organized version history

XuanYang-cn:
@s-gruneberg Overall LGTM to me.
/lgtm
/approve

XuanYang-cn (approved):

PR #214, opened 2023-08-18
Wait for index done

XuanYang-cn:
image

### DIFF
--- a/vectordb_bench/backend/clients/milvus/milvus.py
+++ b/vectordb_bench/backend/clients/milvus/milvus.py
@@ -1,6 +1,7 @@
 """Wrapper around the Milvus vector database over VectorDB"""
 
 import logging
+import time
 from contextlib import contextmanager
 from typing import Iterable, Type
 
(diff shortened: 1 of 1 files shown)


liliu-z (approved):

XuanYang-cn:
@liliu-z This should do it for now
image

agsachin:
@XuanYang-cn @liliu-z can you please explain what is the problem which is addressed with this code change 

1. I can see that you are creating the index before inserting the data? what did we achieve by doing this ?
2. we are already waiting for `wait_for_index_building_complete` then why are we again checking for `wait_index`
3. I can see you have moved the compact() after indexing ? again what did we achieve by doing this ? 

```
    utility.wait_for_index_building_complete(self.collection_name)
            def wait_index():
                while True:
                    progress = utility.index_building_progress(self.collection_name)
                    if progress.get("pending_index_rows", -1) == 0:
                        break
                    time.sleep(5)

            wait_index()
            self.col.compact()
            self.col.wait_for_compaction_completed()
            wait_index()
```

Issue #228, opened 2023-10-20
Redis config is drastically wrong for benchmarks


Spartee:
The redis configuration doesn't even use the ANN algorithm ``HNSW``. You're comparing Redis KNN to others ANN. Hopefully just an oversight that Redis supports both. 

**please note**: Anyone who comes to this repo and uses the default configs should be aware of this.

xiaofan-luan:
> The redis configuration doesn't even use the ANN algorithm `HNSW`. You're comparing Redis KNN to others ANN. Hopefully just an oversight that Redis supports both.
> 
> **please note**: Anyone who comes to this repo and uses the default configs should be aware of this.

@Spartee 
could you contribute for the right config? we are glad to take contribution and rerun the test

xiaofan-luan:
<img width="642" alt="image" src="https://github.com/zilliztech/VectorDBBench/assets/83447078/abada480-6dc8-430b-a1a8-d58dc7881840">
 
 I think we need to fallback to HNSW?

Spartee:
Happy to. Working on it already. We will do this.

PR #247, opened 2023-12-01
Send password, batch pipeline

ewpreston:
Set default index type to HNSW instead of FLAT so comparisons are fair

Send password when creating connection. The UI requires a password for Redis. Even if no password is set on the Redis database, sending an empty string is fine.

Batch the pipeline.execute calls so we don't run out of memory on the client.

This is for issue #228 https://github.com/zilliztech/VectorDBBench/issues/228

### DIFF
--- a/vectordb_bench/backend/clients/redis/redis.py
+++ b/vectordb_bench/backend/clients/redis/redis.py
@@ -103,8 +103,7 @@ class Redis(VectorDB):
         Should call self.init() first.
         """
 
-        batch_size = 100 # Adjust this as needed, but don't make too big
-        result_len = 0
+        batch_size = 1000 # Adjust this as needed, but don't make too big
         try:
             with self.conn.pipeline(transaction=False) as pipe:
                 for i, embedding in enumerate(embeddings):
@@ -117,10 +116,9 @@ class Redis(VectorDB):
                     # Execute the pipe so we don't keep too much in memory at once
                     if i % batch_size == 0:
                         res = pipe.execute()
-                        result_len += len(res)
 
                 res = pipe.execute()
-                result_len += len(res)
+                result_len = i + 1
         except Exception as e:
             return 0, e
         
--- a/vectordb_bench/backend/clients/redis/redis.py
+++ b/vectordb_bench/backend/clients/redis/redis.py
@@ -29,8 +29,8 @@ class Redis(VectorDB):
         self.collection_name = INDEX_NAME
 
         # Create a redis connection, if db has password configured, add it to the connection here and in init():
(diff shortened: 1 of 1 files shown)


vectordb_bench/backend/clients/redis/redis.py
```diff
@@ -103,20 +102,29 @@ def insert_embeddings(
         """Insert embeddings into the database.
         Should call self.init() first.
         """
+
+        batch_size = 100 # Adjust this as needed, but don't make too big
```
tylerhutcherson, line 106: Let's start with a slightly more aggressive value which should be beneficial for throughput.. say 1000!

tylerhutcherson (commented):
nice! Eric I left a few small notes

alwayslove2013:
@tylerhutcherson Could you please help review it ~

vectordb_bench/backend/clients/redis/redis.py
```diff
@@ -103,20 +102,27 @@ def insert_embeddings(
         """Insert embeddings into the database.
         Should call self.init() first.
         """
+
+        batch_size = 1000 # Adjust this as needed, but don't make too big
```
tylerhutcherson, line 106: Super minor nit pick, but I think the `i` param is all you need here:

```python
batch_size = 1000 # Adjust this as needed, but don't make too big
try:
    with self.conn.pipeline(transaction=False) as pipe:
        for i, embedding in enumerate(embeddings, start=1):
            embedding = np.array(embedding).astype(np.float32)
            pipe.hset(metadata[i], mapping = {
                "id": str(metadata[i]),
                "metadata": metadata[i], 
                "vector": embedding.tobytes(),
            })
            # Execute the pipe so we don't keep too much in memory at once
            if i % batch_size == 0:
                res = pipe.execute()

        res = pipe.execute()
except Exception as e:
    return 0, e

return i, None

```

tylerhutcherson:
> @tylerhutcherson Could you please help review it ~

Left one more small suggestion for @ewpreston but otherwise LGTM

vectordb_bench/backend/clients/redis/redis.py
ewpreston: If I start i =1 then i ends with 5001 instead of 5000. That results in an index out of bounds error by the caller.

tylerhutcherson (approved):
LGTM! Left one comment, but @ewpreston corrected me, so this should be good to go.

alwayslove2013 (approved):

alwayslove2013:
@ewpreston @tylerhutcherson nice~

PR #319, opened 2024-05-08
Optimize pgvector test for semi-recent enhancements

jkatz:
This commit adds several changes to the pgvector test to create a more representative test environment based on recent and older changes to pgvector. Notable changes include allowing for testing of parallel index buiding parameters, using loading with the recommended binary loading method, and other changes to better emulate what a typical user of pgvector would do.

This commit also has some general cleanups as well.

Co-authored-by: Mark Greenhalgh <greenhal@users.noreply.github.com>
Co-authored-by: Tyler House <tahouse@users.noreply.github.com>

### DIFF
--- a/vectordb_bench/backend/clients/pgvector/config.py
+++ b/vectordb_bench/backend/clients/pgvector/config.py
@@ -1,29 +1,62 @@
+from abc import abstractmethod
+from typing import Any, Mapping, Optional, Sequence, TypedDict
 from pydantic import BaseModel, SecretStr
-from ..api import DBConfig, DBCaseConfig, IndexType, MetricType
+from typing_extensions import LiteralString
+from ..api import DBCaseConfig, DBConfig, IndexType, MetricType
 
 POSTGRE_URL_PLACEHOLDER = "postgresql://%s:%s@%s/%s"
 
+
+class PgVectorConfigDict(TypedDict):
+    """These keys will be directly used as kwargs in psycopg connection string,
+        so the names must match exactly psycopg API"""
+
+    user: str
+    password: str
+    host: str
+    port: int
+    dbname: str
+
+
 class PgVectorConfig(DBConfig):
-    user_name: SecretStr = "postgres"
+    user_name: SecretStr = SecretStr("postgres")
     password: SecretStr
     host: str = "localhost"
     port: int = 5432
     db_name: str
 
-    def to_dict(self) -> dict:
+    def to_dict(self) -> PgVectorConfigDict:
         user_str = self.user_name.get_secret_value()
         pwd_str = self.password.get_secret_value()
         return {
-            "host" : self.host,
-            "port" : self.port,
-            "dbname" : self.db_name,
-            "user" : user_str,
-            "password" : pwd_str
+            "host": self.host,
+            "port": self.port,
+            "dbname": self.db_name,
+            "user": user_str,
+            "password": pwd_str,
         }
+
+
+class PgVectorIndexParam(TypedDict):
+    metric: str
+    index_type: str
+    index_creation_with_options: Sequence[dict[str, Any]]
+    maintenance_work_mem: Optional[str]
+    max_parallel_workers: Optional[int]
+
+
+class PgVectorSearchParam(TypedDict):
+    metric_fun_op: LiteralString
+
+
+class PgVectorSessionCommands(TypedDict):
+    session_options: Sequence[dict[str, Any]]
+
 
 class PgVectorIndexConfig(BaseModel, DBCaseConfig):
     metric_type: MetricType | None = None
-    index: IndexType
+    create_index_before_load: bool = False
+    create_index_after_load: bool = True
 
     def parse_metric(self) -> str:
         if self.metric_type == MetricType.L2:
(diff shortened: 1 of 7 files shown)


jkatz:
@XuanYang-cn @alwayslove2013 Please let us know if this PR requires additional work. There are some other changes we'd like to include for testing other configurations of pgvector, but we'd like to baseline it against the flat implementation first. Thanks!

vectordb_bench/backend/clients/pgvector/config.py
```diff
@@ -1,29 +1,64 @@
+from abc import abstractmethod
+from typing import Any, Mapping, Optional, Sequence, TypedDict
+
+from psycopg import sql
```
alwayslove2013, line 4: Recommend avoiding adding specific dependencies in the `config.py`. Users **only install the corresponding toolkit when they are conducting tests.**

However, in the scenario where the default results page is opened solely for result display, `VDBBench` will load the standardized result (json). Serializing this data requires the `config.py` file from all clients.

Currently, if a user hasn't installed `psycopg`, they won't be able to open the results page.

jkatz:
@alwayslove2013 Thanks for the feedback! This is resolved in the latest push.

Overall, I would suggest moving to psycopg3 (`psycopg`) as it's now the maintained version of psycopg; however, that change could be made in a separate pull request.

vectordb_bench/backend/clients/pgvector/config.py
jkatz: This is fixed in ea29f47

vectordb_bench/backend/clients/pgvector/config.py
```diff
@@ -46,48 +79,142 @@ def parse_metric_fun_str(self) -> str:
...
+    ) -> Sequence[dict[str, Any]]:
+        """Walk through options, creating 'SET 'key1 = "value1";' commands"""
+        session_options = []
+=======
+    ) -> Sequence[dict[str, Any]]:
+        session_options = []
+>>>>>>> 1a4f510 (Remove import of psycopg from config.)
+        for setting_name, value in set_mapping.items():
+            if value:
+                session_options.append(
```
alwayslove2013, line 121: @jkatz need resolve.

alwayslove2013 (changes requested):

alwayslove2013:
@jkatz Thank you so much for your contribution! We greatly appreciate it and are thrilled to receive your pull request. We look forward to collaborating with you and driving the project forward together!

> Overall, I would suggest moving to psycopg3 (psycopg) as it's now the maintained version of psycopg; however, that change could be made in a separate pull request.

vectordb_bench/backend/clients/pgvector/config.py
jkatz: 🤦 That's embarrassing. I'll fix that on the next pass.

jkatz: For the record sake - I had previously resolved a merge conflict, but for some reason it left the merge flags in. That is likely my misread from VSCode. The next commit fixes this.

jkatz:
@alwayslove2013 Likewise. I personally appreciate the approach VectorDBBench takes around testing concurrency, which resembles how users interact with databases.

I've pushed up the fix to the latest patch to handle the merge conflict that remained (which I'm still baffled how that got in, but I'll triple check next time).

alwayslove2013 (approved):
/approve

alwayslove2013:
@jkatz I would like to express my sincere gratitude for your support. Our primary goal has been to ensure that the test data reflects the performance characteristics of the real-world usage scenarios as accurately as possible.

If you have any suggestions or innovative ideas with VDBBench, we would be more than happy to discuss them with you. Your valuable input is crucial for us to enhance the functionality and user experience of the tool.

wahajali:
@jkatz I'm trying to see how we can optimize the pgvector and pgvecto.rs client further and one thought I had was to use stored procedures for making vector queries. Do you think its something to explore further? 
I've previously seen something similar for HammerDB (although its not for vector benchmarking): https://www.hammerdb.com/blog/uncategorized/why-you-should-benchmark-your-database-using-stored-procedures/

greenhal:
This benchmark is pretty simple, run the query, get the result, compare the result.  

To improve the performance, using stored procedures you would have to either reduce network round trips, or reduce the data transmitted, neither of which is much overhead in this benchmark.  Using stored procedures HammerDB does both of these, because there is a lot of potential network traffic associated with a tpcc benchmark, for example see [new order transactions](https://courses.cs.washington.edu/courses/csep545/01wi/lectures/class1/tsld039.htm),  to create a new order (the main tpcc performance measurement), it could take up to 6 round trips to create an order, if all of the logic was on the client, using stored procedures in this case, it's 1  or even less, because hammer db can send a single request that says, create 100 orders. 


With that in mind, there could be 2 ways that could possibly improve pgvector performance, or any engines performance.  (which could be implemented with stored procedures.)

1) Reduce the network trips.  
This would be accomplished by batching the request/results, instead of sending 1 vector to query, send 100 or even all 1k, then return all of the results.  But since the same amount of data would be sent/received. The only savings would be the reduce round trips and the overhead associated with those round trips and the sending/receiving would still be done over multiple trips.   Overall, the benefit would be minimal, if any at all.      

2) Reduce the data transferred.   
There are a few ways of doing this,  the most extreme would be: 
   - to load the data, ground truth and test vector onto the target
   - a "run benchmark" procedure on the target would run all the queries and just return the recall and qps.

Both would require a pretty significant change to vectordbbench, they would be engine specific and would not represent a real world use case.

PR #366, opened 2024-08-30
Add quantization option for pgvector with support for halfvec

lucagiac81:
This PR adds a quantization option for pgvector.

The option is _quantization_type_, following the same naming as for pgvecto_rs. The option is added to both the web UI and CLI. 

The supported values are _none_ and _halfvec_.
- For the CLI, not specifying the option is equivalent to selecting _none_.
- When _halfvec_ is selected, scalar FP16 quantization is applied when creating the index (HNSW or IVFFlat).
- Additional quantization types can be added in the future (e.g., bit vectors).

### DIFF
--- a/README.md
+++ b/README.md
@@ -117,6 +117,8 @@ Options:
   --m INTEGER                     hnsw m
   --ef-construction INTEGER       hnsw ef-construction
   --ef-search INTEGER             hnsw ef-search
+  --quantization-type [none|halfvec]
+                                  quantization type for vectors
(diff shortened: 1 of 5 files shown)


lucagiac81:
Resolved conflicts due to recently merged PR #362

alwayslove2013 (approved):

jkatz:
FWIW I think this is a good start - we should add in binary quantization support as well. However, because BQ requires reranking, there may be some additional work to do here.

PR #389, opened 2024-10-28
Binary Quantization Support for pgvector HNSW Algorithm

Sheharyar570:
This update introduces binary vector quantization support to the `pgvector` HNSW algorithm within our benchmarking tool. The implementation includes binary vector quantization queries with an option for reranking.
### New CLI and UI Options
- **`reranking`** *(boolean)*: Enables or disables reranking for quantized queries.
- **`quantized_fetch_limit`** *(integer)*: Specifies the number of rows to fetch for distance calculations between binary vectors before reranking.
- **`reranking_metric`** *(string)*: Chooses the distance metric for reranking, with options:
  - `COSINE`
  - `L2`
  - `IP`


greenhal:
This is a great addition to VectorDBBench.  My only suggestion is would be to try to reduce the large section of duplicate code in pgvector with the filtered/unfiltered sql query definition. Lines 132-194 & 195-245, this code is identical except for the where clause.

vectordb_bench/frontend/components/run_test/caseSelector.py
```diff
@@ -103,10 +103,12 @@ def caseConfigSetting(st, dbToCaseClusterConfigs, uiCaseItem: UICaseItem, active
                 elif config.inputType == InputType.Float:
                     caseConfig[config.label] = column.number_input(
                         config.displayLabel if config.displayLabel else config.label.value,
-                        step=config.inputConfig.get("step", 0.1),
-                        min_value=config.inputConfig["min"],
-                        max_value=config.inputConfig["max"],
-                        key=key,
+                        value=config.inputConfig["value"],
+                        help=config.inputHelp,
+                    )
```
alwayslove2013, line 108: Could you please share if there is any specific reason for considering the removal of "min" / "max" / "step" / "key"?

Sheharyar570: Thanks for pointing out, I made a mistake here while adding a new input type. I've reverted these changes. Can you please review it again.

alwayslove2013 (approved):

PR #390, opened 2024-10-29
increase timeout

alwayslove2013:
Increase the timeout limit significantly to reduce the risk of re-testing for first-time users, while allowing advanced users the option to set this parameter. 

In the future, consideration can be given to exposing the timeout parameter in the web UI.


XuanYang-cn (approved):

liliu-z (commented):
@alwayslove2013 @XuanYang-cn Let's make it tunable from the UI. We can only put the results satisfy the default value on the leaderboard.

PR #410, opened 2024-11-25
fix conc_latency_p99 calculation; add conc_latency_avg metric; conc_test first

alwayslove2013:
- fix conc_latency_p99 calculation
- add conc_latency_avg metric
- do conc_test first, then serial_test, to ensure the database is warm during a serial test.


XuanYang-cn (approved):
/lgtm

PR #427, opened 2024-12-16
Add table quantization option for pgvector

lucagiac81:
This PR adds a table quantization option for pgvector. 
- The _table_quantization_type_ option controls quantization of data as it is loaded into the database. 
- This is in addition to the existing _quantization_type_ option which controls quantization of data in an index.

Quantizing data as it is loaded enables storage savings if the original float vectors do not need to be preserved for other purposes. For example, halfvec quantization could be used in both the table and index, reducing the size of the table and eliminating the need for conversions.

_table_quantization_type_ supports the same quantization options as _quantization type_: _none, halfvec, bit._ 
Most combinations of the two options are supported, with some exceptions. 
- If _table_quantization_type_ = _bit_, then _quantization type_ is forced to _bit_ as well, as it does not make sense to convert bit vectors in the table to vector or halfvec in an index.
- Reranking is only supported if _quantization_type_ = _bit_ and _table_quantization_type_ is vector or halfvec.

### DIFF
--- a/README.md
+++ b/README.md
@@ -131,7 +131,11 @@ Options:
   --ef-construction INTEGER       hnsw ef-construction
   --ef-search INTEGER             hnsw ef-search
   --quantization-type [none|bit|halfvec]
-                                  quantization type for vectors
+                                  quantization type for vectors (in index)
+  --table-quantization-type [none|bit|halfvec]
+                                  quantization type for vectors (in table). If
+                                  equal to bit, the parameter
+                                  quantization_type will be set to bit too.
   --custom-case-name TEXT         Custom case name i.e. PerformanceCase1536D50K
   --custom-case-description TEXT  Custom name description
   --custom-case-load-timeout INTEGER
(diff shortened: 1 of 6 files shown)


alwayslove2013 (approved):

akashsha1 (approved):

lucagiac81:
Formatting checks are failing for files not modified by this PR. Would you like me to fix them in this PR? If they are fixed in a separate PR, I will rebase on it.

```
PYTHONPATH=`pwd` python3 -m black vectordb_bench --check
would reformat /home/runner/work/VectorDBBench/VectorDBBench/vectordb_bench/backend/clients/milvus/config.py
would reformat /home/runner/work/VectorDBBench/VectorDBBench/vectordb_bench/backend/clients/milvus/cli.py
```

alwayslove2013 (approved):

alwayslove2013:
@lucagiac81 I will check it.
> Formatting checks are failing for files not modified by this PR.

PR #434, opened 2025-01-07
Removed the Filter Path from Opensearch Search, so we can get the full response from search

Xavierantony1982:
This is a fix for the search response bug for opensearch, Search is not returning the shards and hits total in the response due to the filter.

One line change to remove the filter from response.


navneet1v (approved):

alwayslove2013 (approved):

navneet1v:
Thanks @alwayslove2013 for merging the change

PR #480, opened 2025-02-27
custom control file and column name

Caroline-an777:


vectordb_bench/backend/dataset.py
```diff
@@ -106,6 +113,20 @@ def dir_name(self) -> str:
     def file_count(self) -> int:
         return self.file_num
 
+    @property
+    def train_files(self) -> list[str]:
+        train_file, train_count = self.train_file, self.file_count
+        prefix = f"{train_file}"
+        train_files = []
+        if train_count > 1:
+            prefix_s = prefix.split(",")
```
alwayslove2013, line 122: Using `.strip()` to handle the spaces around the delimiter would be better.

alwayslove2013 (approved):

PR #484, opened 2025-03-12
Add TiDB backend

breezewish:
This PR adds TiDB as an available backend, which is previously maintained in our own fork: https://github.com/wd0517/VectorDBBench. As now the vector index syntax is formalized and will not change, I think it's a good time to contribute it to the community 👏 

- When connecting to a local cluster, usually SSL need to set to False. When connecting to TiDB Serverless, SSL must set to True.
- Currently there is no index parameters available for adjustments and the index building is required.
- Filter is currently not implemented in this PR. It will be added in the next PR.

UI:

image

CLI sample:

```shell
vectordbbench tidb --case-type Performance768D1M --num-concurrency 30
```

image

image

### DIFF
--- a/README.md
+++ b/README.md
@@ -55,6 +55,7 @@ All the database client supported
 | awsopensearch            | `pip install vectordb-bench[opensearch]` |
 | aliyun_opensearch        | `pip install vectordb-bench[aliyun_opensearch]` |
 | mongodb                  | `pip install vectordb-bench[mongodb]`       |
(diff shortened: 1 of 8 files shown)


alwayslove2013:
@breezewish Thanks for your valuable contribution to `VectorDBBench`! It will significantly improve our community. We appreciate your hard work and dedication.

alwayslove2013 (approved):

Issue #489, opened 2025-03-25
NameError: name 'MilvusGPUBruteForceTypedDict' is not defined


pratim808:
When i run this command - "vectordbbench --help" i am getting this error 

Traceback (most recent call last):
  File "/home/manas/Downloads/vectordbbench/new-vectordbbench/vdb_env/bin/vectordbbench", line 5, in <module>
    from vectordb_bench.cli.vectordbbench import cli
  File "/home/manas/Downloads/vectordbbench/new-vectordbbench/vdb_env/lib/python3.12/site-packages/vectordb_bench/cli/vectordbbench.py", line 5, in <module>
    from ..backend.clients.milvus.cli import MilvusAutoIndex
  File "/home/manas/Downloads/vectordbbench/new-vectordbbench/vdb_env/lib/python3.12/site-packages/vectordb_bench/backend/clients/milvus/cli.py", line 198, in <module>
    @click_parameter_decorators_from_typed_dict(MilvusGPUBruteForceTypedDict)
                                                ^^^^^^^^^^^^^^^^^^^^^^^^^^^^
NameError: name 'MilvusGPUBruteForceTypedDict' is not defined

how can i solve this?

yuyuankang:
I am faced with the same issue this morning. I addressed it by removing some duplicated codes from `cli.py`. I just submitted PR #490.

PR #490, opened 2025-03-25
remove duplicated code

yuyuankang:
This pull request is related to #489

I found some codes are duplicated. Removing these lines of code makes `vectordbbench --help` work correctly.


alwayslove2013 (approved):
Nice!

PR #518, opened 2025-05-06
Add lancedb support

s-h-a-d-o-w:
Pretty straightforward, except that I've only been able to run with HSNW once on my machine. I don't know what's going on with that.
The linting exception that I had to add is the one for not allowing more than 50 statements.

(I have changed the command block bits in the readme because I found it a bit annoying that I wasn't able to just copy the commands. 😅 )


alwayslove2013 (approved):

alwayslove2013:
@s-h-a-d-o-w We really appreciate your efforts for `LanceDB` support and believe it will benefit our community greatly.

PR #521, opened 2025-05-12
Add --concurrency-timeout option to avoid long time waiting

LoveYou3000:
Add --concurrency-timeout option to avoid long time waiting, by default, it's 3600s.

### DIFF
--- a/README.md
+++ b/README.md
@@ -114,6 +114,10 @@ Options:
   --num-concurrency TEXT          Comma-separated list of concurrency values
                                   to test during concurrent search  [default:
                                   1,10,20]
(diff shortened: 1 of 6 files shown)


alwayslove2013 (approved):

alwayslove2013:
@LoveYou3000 I think this PR is great and believe it will help us identify the issue of "subprocess creation failure leading to system freeze" due to insufficient computational resources. 

However, I'm still a bit uncertain whether the system can reschedule to the error-raising part after it has completely frozen due to **resource shortages**. Perhaps we need a safer method for creating subprocesses that stops the process creation and raises an exception at the system level when resource constraints are detected.

PR #523, opened 2025-05-13
Fixing Bugs in Benchmarking ClickHouse with vectordbbench

yuyuankang:
Currently, I found three issues when running `vectordbbench` to benchmark a local ClickHouse instance:

1. The `--user <username>` argument in `vectordbbench clickhouse` is not passed to the ClickHouse client, so it always defaults to the username "clickhouse".
1. According to the [ClickHouse documentation](https://clickhouse.com/docs/engines/table-engines/mergetree-family/annindexes), the `vector_similarity` syntax is incorrect — the third parameter should specify the dimension.
1. The benchmark fails when no filter is provided, raising the error: `AttributeError: 'NoneType' object has no attribute 'get'`.

### DIFF
--- a/vectordb_bench/backend/clients/clickhouse/cli.py
+++ b/vectordb_bench/backend/clients/clickhouse/cli.py
@@ -1,66 +1,89 @@
-from typing import Annotated, TypedDict, Unpack
+from abc import abstractmethod
+from typing import TypedDict
 
-import click
-from pydantic import SecretStr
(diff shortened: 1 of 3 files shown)


alwayslove2013 (approved):

alwayslove2013:
@yuyuankang Fantastic~ I believe it will greatly help the community better test `ClickHouse`.

PR #526, opened 2025-05-19
add --num-shards option for milvus performance test case

LoveYou3000:


alwayslove2013:
@LoveYou3000, thanks for your PR, but I'm curious why you wanna change the `num_shards` for Milvus. Given that each `shard` in Milvus can handle at least 200 million data rows and that the public datasets provided by vdbbench are significantly smaller, we did not include an option for `num_shards` when designing the Milvus client.

Are you testing with a much larger dataset, or do you have a specific need for extremely high write speeds? Otherwise, we generally do not recommend increasing the number of shards.

alwayslove2013 (approved):

PR #529, opened 2025-05-25
Fixing bugs in aws opensearch client and added fp16 support 

navneet1v:
Fixing bugs in aws opensearch client and added fp16 support

Testing using

```
vectordbbench awsopensearch --db-label 10m-run --task-label ef-search-40 --m 16 --ef-construction 200 --ef-runtime 100 --host <HOST-Name> --port 443 --user admin --password '<MYpass>' --case-type Performance768D10M --number-of-shards 1  --number-of-replicas 0   --num-concurrency 10,30,40,150,200 --k 10 --quantization-type fp16

```

### DIFF
--- a/README.md
+++ b/README.md
@@ -201,9 +201,12 @@ Options:
   --force-merge-enabled BOOLEAN   Whether to perform force merge operation
   --flush-threshold-size TEXT     Size threshold for flushing the transaction
                                   log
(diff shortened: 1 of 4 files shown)


navneet1v:
@alwayslove2013 all the CIs passed, after the fix you suggested. Thanks. Can we merge this change now?

PR #537, opened 2025-06-10
expose lancedb index parameters to the cli interface

ZebinRen:
Expose the lancedb index parameters to the cli interface. Omitting these parameters will set the same value as the previous version, so this pull request will not change the behavior of existing scripts. Following is a list of the parameters that can be set with the cli interface:

lancedbivfpq
- num-partitions
- num-sub-vectors
- nbits
- nprobes

lancedbhnsw
- m
- ef-construct
- ef


alwayslove2013:
@ZebinRen, there are some style issues. could you help check it~
```sh
pip install ruff --upgrade
pip install black --upgrade
make format
make lint
```

ZebinRen:
Thanks for the advice! I have run the lint test and fixed the formatting issues.

alwayslove2013 (approved):

PR #556, opened 2025-07-07
Fixed the issue where the welcome page image could not be loaded.

zhuwenxing:
The current method of reading images only works when running in the root directory of the project. If the package is installed via pip and then initialized with init_bench, it cannot be used.


zhuwenxing:
image
current issue

@alwayslove2013

alwayslove2013 (approved):

PR #586, opened 2025-08-19
feat(pgvector): add new label/int-filter support and refactor client

alwayslove2013:


alwayslove2013:
@wahajali could you please help review it~

PR #614, opened 2025-10-15
awsopensearch add normalize for cosine similarity when indexing data

norrishuang:
awsopensearch add normalize for cosine similarity, solved issue of recall rate with Cohere dataset

### DIFF
--- a/vectordb_bench/backend/clients/aws_opensearch/aws_opensearch.py
+++ b/vectordb_bench/backend/clients/aws_opensearch/aws_opensearch.py
@@ -65,11 +65,11 @@ class AWSOpenSearch(VectorDB):
             self._load_graphs_to_memory(client)
 
     def need_normalize_cosine(self) -> bool:
-        if self.case_config.metric_type.upper() == 'COSINE':
+        if self.case_config.metric_type.upper() == "COSINE":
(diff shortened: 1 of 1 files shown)


vectordb_bench/backend/clients/aws_opensearch/aws_opensearch.py
```diff
@@ -64,6 +64,12 @@ def __init__(
             self._update_ef_search_before_search(client)
             self._load_graphs_to_memory(client)
 
+    def need_normalize_cosine(self) -> bool:
+        if self.case_config.metric_type.upper() == "COSINE":
+            log.info("cosine dataset need normalize.")
+            return True
+        return False
```
alwayslove2013: The `need_normalize_cosine` function is designed for databases that do NOT natively support cosine similarity metrics. When testing cosine-based datasets, this function normalizes all training and query vectors before tests, to make the metric equivalent to L2 (Euclidean) or IP (Inner Product), ensuring accurate recall calculations. This is a workaround solution when cosine similarity is not directly supported.

For databases that already support cosine similarity, we do not recommend enabling this feature. The goal is to maintain consistent benchmarking across different databases and ensure fair testing conditions.

Issue #621, opened 2025-10-28
mistakes in README.md for package installation pip commands


towfeeqfayaz11:
##### WHAT
The commands listed on the page when run as it is fails

e.g;
```
pip install vectordb-bench[qdrant]
zsh: no matches found: vectordb-bench[qdrant]
```

PR #622, opened 2025-10-28
fixing README.md file to update pip install command

towfeeqfayaz11:
This is fix for issue: [#621](https://github.com/zilliztech/VectorDBBench/issues/621)


alwayslove2013:
@towfeeqfayaz11 `pip install vectordb-bench[pinecone]` works in most cases. However, in some shells, the characters `[` and `]` are interpreted as wildcards, which can cause unexpected behavior. Adding quotes around it like `pip install 'vectordb-bench[pinecone]'`, will prevent the shell from parsing those special characters and let `pip` handle them correctly instead.

Your change is really necessary and helpful. We truly appreciate your contribution to the project!

PR #647, opened 2025-11-26
fix(cockroachdb): Handle 30s timeout and ensure vector index usage

viragtripathi:
Two critical fixes for multi-node CockroachDB clusters:

1. Connection Timeout Handling: On multi-node v25.4 clusters, CREATE VECTOR INDEX from subprocess contexts experiences a 30-second connection timeout. The index creation continues successfully in the background. This fix detects the timeout and polls for completion (up to 5 minutes).

2. Vector Index Usage: Fixed vector_search_beam_size not being set on pooled connections, causing queries to use full table scan instead of the vector index. Now configures every connection from the pool with proper beam size.

Testing:
- Single-node: Works without timeout (178s index creation)
- Multi-node: Successfully handles timeout and completes (131s total)
- Vector index: Now properly used for all searches (verified with EXPLAIN)
- Both achieve ~83% recall with good QPS

Fixes issues where:
- Benchmarks would fail despite successful index creation
- Searches were slow due to full table scans instead of index usage

### DIFF
--- a/vectordb_bench/backend/clients/cockroachdb/cockroachdb.py
+++ b/vectordb_bench/backend/clients/cockroachdb/cockroachdb.py
@@ -35,7 +35,7 @@ class CockroachDB(VectorDB):
         FilterOp.StrEqual,
     ]
 
-    def __init__(
+    def __init__(  # noqa: PLR0915
         self,
         dim: int,
         db_config: dict,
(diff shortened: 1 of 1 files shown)


viragtripathi:
@alwayslove2013 Please take a look and merge this fix/PR. Thank you!

alwayslove2013 (approved):

PR #652, opened 2025-11-27
cli for elastic by douwei.

Denise2004:
cli for elastic. The program has implemented both filtering mode and non-filtering mode. The command will eventually generate a JSON file, but the filtering rate has not yet been set with default parameters.


vectordb_bench/config-files/elastic.yml
```diff
@@ -0,0 +1,115 @@
+elasticcloudhnsw:
```
alwayslove2013: should remove

alwayslove2013 (changes requested):

PR #654, opened 2025-11-28
Add Tencent Elasticsearch Results

morning-color:


vectordb_bench/results/TencentElasticsearch/result_20251128_2025112810_tencentelasticsearch.json
```diff
@@ -0,0 +1,135 @@
...
+          "refresh_interval": "-1",
+          "merge_max_thread_count": 8,
+          "use_rescore": false,
+          "oversample_ratio": 2,
+          "use_routing": false,
+          "use_force_merge": true,
+          "metric_type": "COSINE",
+          "efConstruction": 200,
+          "M": 16,
+          "num_candidates": 4000
```
alwayslove2013: Thank you for sharing your test results on `Tencent-ES` with us. However, the `num_candidates` parameter appears to be set higher than typical. For tasks targeting `top-100` retrieval, it is common to set this value around `100`, which can achieve `recall=0.9`. Here, setting it to `4000` achieves a recall above 0.99, but it also significantly increases computational overhead. Please confirm whether this meets your expectations.

PR #671, opened 2025-12-08
make collection name configurable

lantingchiang:


vectordb_bench/backend/clients/zilliz_cloud/config.py
```diff
@@ -9,13 +9,15 @@ class ZillizCloudConfig(DBConfig):
...
     num_shards: int = 1
+    collection_name: str = "ZillizCloudVDBBench"
 
     def to_dict(self) -> dict:
         return {
             "uri": self.uri.get_secret_value(),
             "user": self.user,
             "password": self.password.get_secret_value(),
             "num_shards": self.num_shards,
+            "collection_name": self.collection_name,
```
alwayslove2013: looks like the current changes are still not sufficient to take effect. Ultimately, it will require adjustments up to this line. https://github.com/zilliztech/VectorDBBench/blob/21eafc102e1526693058c1721fd68a4f71b8a3be/vectordb_bench/backend/clients/milvus/milvus.py#L42

alwayslove2013 (changes requested):

alwayslove2013 (approved):

PR #682, opened 2025-12-23
feat(milvus): add SCANN index support

JackLCL:
Issue: https://github.com/zilliztech/VectorDBBench/issues/681

- Add SCANN index type for Milvus
- Build params: nlist (default: 1024), with_raw_data (default: false)
- Search params: reorder_k (default: 100)
- Add UI configuration for SCANN index

### DIFF
--- a/vectordb_bench/backend/clients/api.py
+++ b/vectordb_bench/backend/clients/api.py
@@ -41,6 +41,7 @@ class IndexType(str, Enum):
     GPU_IVF_PQ = "GPU_IVF_PQ"
     GPU_CAGRA = "GPU_CAGRA"
     SCANN = "scann"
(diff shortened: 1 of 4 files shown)


vectordb_bench/backend/clients/api.py
```diff
@@ -40,7 +40,7 @@ class IndexType(str, Enum):
     GPU_BRUTE_FORCE = "GPU_BRUTE_FORCE"
     GPU_IVF_PQ = "GPU_IVF_PQ"
     GPU_CAGRA = "GPU_CAGRA"
-    SCANN = "scann"
+    SCANN = "SCANN"
```
alwayslove2013: Do not overwrite existing types. `SCANN_MILVUS` would be better.

alwayslove2013 (approved):

PR #686, opened 2025-12-24
feat(milvus): add nprobe parameter for SCANN index

JackLCL:
- Add nprobe search parameter for SCANN index (default: 64)
- Enable nprobe UI configuration for SCANN_MILVUS index type


chasingegg (approved):
/lgtm

PR #724, opened 2026-02-13
Feat: Add label filter support in pgdiskann client

EeshaaKhan:
This PR adds support for label-based filtering in pgdiskann client.

Changes introduced:

- Added support for FilterOp.StrEqual to enable filtering by scalar string labels.
- Introduced optional with_scalar_labels flag to create tables with an additional label column.
- Updated table creation logic to conditionally include scalar label support.
- Extended insert_embeddings to handle label data when enabled.
- Implemented prepare_filter() to dynamically construct the WHERE clause based on filter type.
- Unified search query generation through _generate_search_query() to support filtered and unfiltered search paths.
- Existing numeric filtering (id >= value) and non-filtered search behavior remain unchanged.
- Includes vector storage optimization (`SET STORAGE PLAIN`) to disable TOAST compression on the embedding column for improved query performance

Screenshot from 2026-02-13 11-53-39

### DIFF
--- a/vectordb_bench/backend/clients/pgdiskann/pgdiskann.py
+++ b/vectordb_bench/backend/clients/pgdiskann/pgdiskann.py
@@ -21,16 +21,15 @@ class PgDiskANN(VectorDB):
     """Use psycopg instructions"""
 
     supported_filter_types: list[FilterOp] = [
-            FilterOp.NonFilter,
-            FilterOp.NumGE,
-            FilterOp.StrEqual,  
+        FilterOp.NonFilter,
+        FilterOp.NumGE,
+        FilterOp.StrEqual,
     ]
 
     conn: psycopg.Connection[Any] | None = None
     cursor: psycopg.Cursor[Any] | None = None
 
-    _filtered_search: sql.Composed
-    _unfiltered_search: sql.Composed
+    _search: sql.Composed  
 
     def __init__(
         self,
@@ -49,8 +48,7 @@ class PgDiskANN(VectorDB):
         self.dim = dim
         self.with_scalar_labels = with_scalar_labels
         self._scalar_label_field = "label"
-        self.filter_op = None
-        self.filter_value = None
+        self.where_clause = "" 
 
         self._index_name = "pgdiskann_index"
         self._primary_field = "id"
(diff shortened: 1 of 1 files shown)


EeshaaKhan:
@XuanYang-cn Could you please review this PR?

EeshaaKhan:
@alwayslove2013 Hi, this PR has been pending review for a while. Could you please take a look?

XuanYang-cn:
Hi @EeshaaKhan, thanks for the contribution, and apologies for the delayed response! I'll review this PR shortly.

Before I do, could you please rebase onto the latest main? We've since upgraded to Pydantic v2 and updated several checks, so this PR didn't trigger the GitHub Actions workflow and may fail against the current codebase.
Thanks again for your patience!

EeshaaKhan:
Hi @XuanYang-cn ,
Thank you for the feedback! I've rebased the PR onto the latest main branch. All checks are now passing with no conflicts.
Ready for review! 🚀

XuanYang-cn (approved):

XuanYang-cn (commented):
Thanks for the PR! The overall approach is clean and consistent with how pgvector and cockroachdb handle label filtering.

**Worth noting**

- **`ALTER TABLE ... SET STORAGE PLAIN` runs unconditionally in `_create_table`**, regardless of `with_scalar_labels`. Disabling TOAST compression affects storage and performance for every PgDiskANN benchmark — worth a brief comment in the code explaining why it's needed, and a mention in the PR description.
- `self._scalar_label_field = "label"` (singular) while `LabelFilter.label_field` defaults to `"labels"` (plural). A short comment on the column name choice would prevent future confusion.

EeshaaKhan:
@XuanYang-cn  Thanks for the feedback!

I've addressed both points:

1. **SET STORAGE PLAIN:** This runs unconditionally because it only targets the vector column - disabling TOAST compression on it is a standard performance optimization for PgDiskANN benchmarks, consistent with how pgvector handles it.

2. **Column naming:** `label` is the column name in the database table, following the same convention used across all other database clients in this codebase. `labels` is the dataset field name. The mismatch is intentional to stay consistent with the existing standard.

PR #760, opened 2026-04-16
fix(pgvector): normalize index_type to lowercase in _create_index to …

shaohuasong-fang:
…match PostgreSQL access method names

PostgreSQL pgvector extension registers index access methods in lowercase (e.g. "hnsw", "ivfflat"), but the frontend passes IndexType.HNSW.value which is uppercase "HNSW", causing "access method HNSW does not exist" error.

PR #762, opened 2026-04-18
Fix: Use sql.SQL() for pgvector access method names to fix "HNSW does not exist" error

### LINKED ISSUE #759: Termial output: "access method 'HNSW' does not exist" Error
When running the VectorDB benchmark (VectorDBBench) against PgVector on webui, the test case fails with the error:
access method "HNSW" does not exist
This occurs during the performance test for the 1M dataset with 768-dimensional vectors, specifically when the benchmark attempts to optimize (create) an HNSW index. The error results in the termination of the benchmark run.
1. VDBB Version:1.0.20
`(.venv) [root@server VectorDBBench-1.0.20]# `
2. PostgreSQL Version
`
[postgres@server ~]$ psql 
psql (18.3)
Type "help" for help.

postgres=# show server_version;
 server_version 
----------------
 18.3
postgres=# show search_path ;
   search_path   
-----------------
 "$user", public
(1 row)
postgres=# SELECT n.nspname as schema_name, p.proname as function_name 
FROM pg_proc p 
JOIN pg_namespace n ON p.pronamespace = n.oid 
WHERE p.proname = 'hnswhandler';
 schema_name | function_name 
-------------+---------------
 public      | hnswhandler
(1 row)

postgres=# SELECT amname FROM pg_am;
 amname  
---------
 heap
 btree
 hash
 gist
 gin
 spgist
 brin
 ivfflat
 hnsw
(9 rows)
`
3. Configuration information

Image

4.Terminal display details:

Image

NagarajuReddyBoggala:
## Problem
PGVector HNSW index creation fails with:
- "access method 'HNSW' does not exist"

This happens through the WebUI but works via CLI.

## Root Cause
In `pgvector.py` lines 363 and 378, the index_type is wrapped with `sql.Identifier()`, which PostgreSQL treats as a quoted identifier. PostgreSQL access methods must NOT be quoted.

## Solution
Changed `sql.Identifier(index_param["index_type"])` to `sql.SQL(index_param["index_type"])`

This generates proper SQL:
- Before: `USING "hnsw"` ❌
- After: `USING hnsw` ✅

## Fixes
Closes #759

### DIFF
--- a/vectordb_bench/backend/clients/pgvector/pgvector.py
+++ b/vectordb_bench/backend/clients/pgvector/pgvector.py
@@ -360,7 +360,7 @@ class PgVector(VectorDB):
                     if index_param["quantization_type"] == "bit"
                     else sql.Identifier("embedding")
                 ),
-                index_type=sql.Identifier(index_param["index_type"]),
+                index_type=sql.SQL(index_param["index_type"]),
                 # This assumes that the quantization_type value matches the quantization function name
                 quantization_type=sql.SQL(index_param["quantization_type"]),
                 dim=self.dim,
@@ -375,7 +375,7 @@ class PgVector(VectorDB):
             ).format(
                 index_name=sql.Identifier(self._index_name),
                 table_name=sql.Identifier(self.table_name),
-                index_type=sql.Identifier(index_param["index_type"]),
+                index_type=sql.SQL(index_param["index_type"]),
                 embedding_metric=sql.Identifier(index_param["metric"]),
             )
 


XuanYang-cn:
@NagarajuReddyBoggala  Thanks for the contribution, please fix the conflict in the pgvector.py so that this fix can be merged safely.

XuanYang-cn:
I think this is a dup with #760 and the #759 was fixed, I'm going to close this PR

@NagarajuReddyBoggala Thanks for the contribution again!

PR #769, opened 2026-04-29
feat(oceanbase): add multi-process loader, configurable index/partition params, and HNSW_BQ cosine support 

wyfanxiao:
Hi team! 👋

This PR enhances the OceanBase client and adds a multi-process data loader to improve load throughput for SQL-based vector DBs.

## Summary
- Add `MultiprocessInsertRunner` for parallel data loading across worker processes, bypassing GIL limitation for SQL-based vector DBs
- OceanBase: declare `thread_safe=False`, auto-switch to multi-process loader when `--load-concurrency>1`
- Add `--load-processes` CLI option for explicit multi-process control
- Add `--create-index-parallel` CLI option (default 16) for configurable index build parallelism
- Add `--extra-info-max-size` CLI option (default 32, set 0 to omit) for HNSW index
- Add `--partitions` CLI option for KEY partitioning (default 0, no partition)
- HNSW_BQ: remove forced L2 for cosine metric, now supports cosine natively
- Improve `concurrent_runner` log message for `thread_safe=False` fallback
- `pyproject.toml`: add `pyyaml` dependency, fix `packages.find` to include all subpackages

We've been using these changes in our OceanBase benchmark testing and they've been working well. Would love to get this merged so it can benefit other users too. Thanks in advance for taking the time to review! Happy to address any feedback. 🙏

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -8,7 +8,7 @@ build-backend = "setuptools.build_meta"
 
 [tool.setuptools.packages.find]
 where = ["."]
-include = ["vectordb_bench", "vectordb_bench.cli"]
+include = ["vectordb_bench", "vectordb_bench.*"]
 
 [project]
 name = "vectordb-bench"
@@ -26,6 +26,7 @@ classifiers = [
 ]
 dependencies = [
     "click",
+    "pyyaml",
     "pytz",
     "streamlit>=1.47,<2",  # 1.47 fixes streamlit#11660
     "tqdm",
--- a/vectordb_bench/backend/clients/oceanbase/cli.py
+++ b/vectordb_bench/backend/clients/oceanbase/cli.py
@@ -31,9 +31,40 @@ class OceanBaseTypedDict(CommonTypedDict):
     ]
     database: Annotated[str, click.option("--database", type=str, help="DataBase name", required=True)]
     port: Annotated[int, click.option("--port", type=int, help="OceanBase port", required=True)]
+    create_index_parallel: Annotated[
+        int,
+        click.option(
+            "--create-index-parallel",
+            type=int,
+            default=16,
+            show_default=True,
+            help="PARALLEL hint degree for CREATE VECTOR INDEX",
+        ),
+    ]
+    partitions: Annotated[
+        int,
+        click.option(
+            "--partitions",
+            type=int,
+            default=0,
+            show_default=True,
+            help="Number of KEY partitions for the table. 0 or 1 means no partitioning.",
+        ),
+    ]
 
 
-class OceanBaseHNSWTypedDict(CommonTypedDict, OceanBaseTypedDict, HNSWFlavor4): ...
+class OceanBaseHNSWTypedDict(CommonTypedDict, OceanBaseTypedDict, HNSWFlavor4):
+    extra_info_max_size: Annotated[
+        int | None,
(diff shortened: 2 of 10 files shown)


wyfanxiao:
Hi @XuanYang-cn. Friendly ping on this PR when you have a moment — happy to adjust based on feedback.

XuanYang-cn (changes requested):
Thanks for the PR. The OceanBase-specific improvements look useful, especially the configurable index parallelism, partitions, extra_info_max_size, and the HNSW_BQ cosine update.

  I do have concerns about merging the shared MultiprocessInsertRunner part as-is, because it changes the benchmark insert procedure rather than only tuning the OceanBase client.

  **My main concern is benchmark correctness and fairness**. For clients that declare
  `thread_safe=False`, this PR auto-switches from the existing single-worker fallback to an N-process loader when `load_concurrency > 1`. That may be a useful direction, but it changes the benchmark contract: some DBs may load through threads, some through one connection, and some through multiple independent processes/connections. We need to be explicit about whether that is considered fair and comparable.

  Also, I don’t think `thread_safe=False` automatically means process-safe. It only tells us one client/connection should not be shared across threads. Multiprocess loading additionally assumes the client object is safely picklable, each child process can reinitialize an independent connection, concurrent inserts into the same table are supported, failures do not leave partial-success benchmark results, and inserted row counts can be verified before optimize/search.

  There are also a few concrete correctness issues in the current runner implementation:

  1. `MultiprocessInsertRunner` can report partial load success. On the normal path it sends sentinels and waits a fixed join window, then terminates workers that are still alive. After that it returns `counter.value` without checking that `inserted == produced`. If the final batches take longer than the join window, the benchmark can continue to optimize/search a partially loaded table.

  2. `load_timeout` is not enforced while the producer is blocked on a full queue. The timeout check only runs after a successful enqueue. If workers hang or become very slow, `_put_with_interruptible_wait` can loop indefinitely as long as at least one worker is still alive.

  3. The timeout path calls `PerformanceTimeoutError(msg)`, but   `PerformanceTimeoutError.__init__` currently takes no message argument. That raises `TypeError` instead of the intended timeout exception, so the case would be marked as a generic failure instead of the existing timeout/out-of-range path.

  **Given this, could we split this PR?**

  My preferred path is:

  1. Keep this PR focused on the OceanBase client/config/CLI changes, which are relatively contained and can likely move forward quickly after smaller review comments are addressed.
  2. Move `MultiprocessInsertRunner`, `--load-processes`, and the auto-switch behavior into a follow-up PR. (I can take the follow-up from there: decide whether multiprocess insert should be  opt-in, auto-enabled only for selected clients, or generalized as the default loader strategy,  and add the tests/metadata needed to preserve benchmark fairness.)

  I’m open to the multiprocess direction, but I think it is larger than the OceanBase client change and needs dedicated verification before becoming part of the shared benchmark execution path.

wyfanxiao:
Hi @XuanYang-cn. Thanks for the thorough review and great suggestions! I've split this into two PRs as you recommended:

- **OceanBase client changes**: #776 (configurable index params, KEY partitioning, HNSW_BQ cosine support)
- **Multi-process loader**: #777 (with all three bugs fixed + fairness discussion)

Happy to continue the discussion on the multi-process PR. Thanks again for the detailed feedback!

PR #770, opened 2026-04-30
feat(seekdb): add SeekDB backend and HNSW benchmark support

liuhao6741:
Summary
-------
Introduce SeekDB as a first-class benchmark target: a MySQL-protocol vector database (OceanBase-style SQL and session variables). The integration supports standard performance workflows and StreamingPerformanceCase-style fixed-rate inserts by combining an upfront HNSW vector index with per-thread connection handling in the rate-based insert runner.

On each benchmark session, SeekDB.init() now applies OceanBase-style tenant system parameters so the engine does not cap workload memory or CPU for the session scope of the connection (ALTER SYSTEM SET memory_limit = "0M" and cpu_count = 0), matching operator guidance for unconstrained resource tests. memory_limit uses the engine-required string size form (e.g. "0M"), not a bare integer.

After bulk load, optimize() runs SELECT VERSION(), parses the embedded seekdb-vX.Y.Z… token (e.g. "5.7.25-OceanBase seekdb-v1.3.0.0"), and when the parsed version is >= 1.3.0 executes CALL dbms_index_manager.refresh() to align index metadata; older or unrecognized version strings skip the call.

New files (vectordb_bench/backend/clients/seekdb/) --------------------------------------------------
- seekdb.py — SeekDB VectorDB implementation
  - Connects via mysql.connector using host/port/user/password/database from SeekDBConfig (Pydantic SecretStr surfaced as plain strings in a TypedDict for the driver).
  - Declares thread_safe=False so ConcurrentInsertRunner does not multiplex multiple threads on one client (mysql.connector connections are not thread-safe). Streaming inserts instead rely on rate_runner’s per-task shallow copy + init() pattern (see below).
  - Lifecycle: __init__ optionally drops/creates table and creates the vector index when drop_old is true; connects temporarily then disconnects.
  - init() context manager reconnects, sets autocommit=1, then applies system parameters for reproducible benchmarking under default tenant limits:
      * ALTER SYSTEM SET memory_limit = "0M"
      * ALTER SYSTEM SET cpu_count = 0 followed, when the case uses HNSW, by search session parameters: * SET ob_hnsw_ef_search=<ef_search> — SeekDB uses OceanBase-style names, not a generic hnsw_ef_search variable. * SET ob_query_timeout=600000000 — raises the default query timeout (microseconds) so ANN queries under concurrent load are less likely to hit short default tenant limits (e.g. 10s).
  - Schema: CREATE TABLE … ORGANIZATION HEAP with INT id and VECTOR(dim) embedding column, as required for vector workloads on this engine.
  - Index: CREATE VECTOR INDEX … WITH (distance=cosine|l2|inner_product, type=HNSW, m=…, ef_construction=…) so the HNSW structure exists before bulk or streaming load (mirrors Milvus-style “index then insert” for concurrent read/write phases).
  - optimize(): SELECT VERSION(), parse seekdb-v… with regex, compare tuple to (1, 3, 0); if >= 1.3.0 run CALL dbms_index_manager.refresh(), else log and return. Requires an active cursor (task_runner calls optimize inside db.init()).
  - insert_embeddings: batched multi-row INSERT with vector literals; default batch size 256 (SEEKDB_DEFAULT_LOAD_BATCH_SIZE).
  - prepare_filter / search_embedding: NonFilter, numeric >= on id, and string equality on id; search uses ORDER BY <metric_func>(embedding, query) APPROXIMATE LIMIT k with metric function names aligned to MetricType (cosine_distance, l2_distance, negative_inner_product).

- config.py — SeekDBConfig (DBConfig), SeekDBHNSWConfig (DBCaseConfig)
  - HNSW index_param() drives CREATE VECTOR INDEX WITH clause.
  - search_param() exposes ef_search for session SET ob_hnsw_ef_search.
  - _seekdb_case_config maps IndexType.HNSW to SeekDBHNSWConfig for DB enum integration.

- cli.py — Click command SeekDBHNSW (registered as seekdbhnsw)
  - SeekDBTypedDict extends CommonTypedDict with --host, --user, --password (default SEEKDB_PASSWORD env or empty), --database, --port.
  - SeekDBHNSWTypedDict adds HNSWFlavor3 (--m, --ef-construction, --ef-search).
  - Invokes shared cli.run() with DB.SeekDB so behavior matches other databases (no SeekDB-specific streaming CLI flags in the shared framework).

Plumbing
--------
- vectordb_bench/backend/clients/__init__.py
  - DB.SeekDB enum value.
  - DB.init_cls branch returning SeekDB client class.
  - DB.config_cls branch returning SeekDBConfig.
  - DB.case_config_cls branch using _seekdb_case_config for HNSW.

- vectordb_bench/cli/vectordbbench.py
  - Import and cli.add_command(SeekDBHNSW).

Rate runner (StreamingPerformanceCase / fixed-rate inserts) -----------------------------------------------------------
- vectordb_bench/backend/runner/rate_runner.py
  - Import shallow copy() in addition to deepcopy().
  - New branch for db.name == "SeekDB": build a thread-local client with copy(db), clear _conn and _cursor so the new object does not share live sockets, then with db_copy.init(): _insert_embeddings(...).
  - Rationale: (1) deepcopy fails or is unsafe on open mysql.connector sockets when worker processes are spawned; (2) sharing one connection across threads violates connector thread-safety and caused streaming insert issues.
  - Other databases keep existing deepcopy or direct-insert behavior.

Documentation for operators
---------------------------
Example invocation (Python 3.11+ recommended for this repo):

  python -m vectordb_bench.cli.vectordbbench seekdbhnsw \
    --case-type StreamingPerformanceCase \
    --host <host> --port 2881 --user root --password '' \
    --database vectordbbench \
    --m 16 --ef-construction 200 --ef-search 64

Ensure the target database exists and mysql-connector-python is installed. The SeekDB user must be allowed to execute ALTER SYSTEM if init-time tuning is required; otherwise connection setup may fail at init().

Notes / non-goals
-----------------
- Only HNSW is registered in _seekdb_case_config; other index types are not wired for SeekDB in this change set.
- No changes to shared cli.py CommonTypedDict or get_custom_case_config for streaming-specific knobs; StreamingPerformanceCase uses framework defaults when launched from CLI.

### DIFF
--- a/vectordb_bench/backend/clients/seekdb/seekdb.py
+++ b/vectordb_bench/backend/clients/seekdb/seekdb.py
@@ -1,0 +1,282 @@
+import logging
+import re
+import struct
+from collections.abc import Generator
+from contextlib import contextmanager
+from typing import Any
+
+import mysql.connector as mysql
+
+from vectordb_bench.backend.filter import Filter, FilterOp
+
+from ..api import IndexType, VectorDB
+from .config import SeekDBConfigDict, SeekDBHNSWConfig
+
+log = logging.getLogger(__name__)
+
+SEEKDB_DEFAULT_LOAD_BATCH_SIZE = 256
+
+# Minimum SeekDB version for dbms_index_manager.refresh() after bulk load (see VERSION()).
+_SEEKDB_REFRESH_MIN_VERSION = (1, 3, 0)
+_SEEKDB_VERSION_IN_VERSION_STRING = re.compile(r"seekdb-v(\d+(?:\.\d+)*)", re.IGNORECASE)
+
+
+def _seekdb_version_tuple(version_row: str | None) -> tuple[int, ...] | None:
+    if not version_row:
+        return None
+    m = _SEEKDB_VERSION_IN_VERSION_STRING.search(version_row.strip())
+    if not m:
+        return None
+    return tuple(int(p) for p in m.group(1).split(".") if p.isdigit())
+
+
+def _version_tuple_ge(parsed: tuple[int, ...], minimum: tuple[int, ...]) -> bool:
+    n = max(len(parsed), len(minimum))
+    for i in range(n):
+        p = parsed[i] if i < len(parsed) else 0
+        m = minimum[i] if i < len(minimum) else 0
+        if p != m:
+            return p > m
+    return True
+
+
+class SeekDB(VectorDB):
+    supported_filter_types: list[FilterOp] = [
+        FilterOp.NonFilter,
+        FilterOp.NumGE,
+    ]
+    # mysql.connector is not thread-safe; ConcurrentInsertRunner uses max_workers=1 when False.
+    # Streaming fixed-rate inserts use rate_runner with per-thread deepcopy+init() instead.
+    thread_safe: bool = False
+
+    def __init__(
+        self,
+        dim: int,
+        db_config: SeekDBConfigDict,
+        db_case_config: SeekDBHNSWConfig,
+        collection_name: str = "items",
+        drop_old: bool = False,
+        **kwargs,
+    ):
+        self.name = "SeekDB"
+        self.dim = dim
+        self.db_config = db_config
+        self.db_case_config = db_case_config
+        self.table_name = collection_name
+        self.load_batch_size = SEEKDB_DEFAULT_LOAD_BATCH_SIZE
+        self._index_name = "vidx"
+        self._primary_field = "id"
+        self._vector_field = "embedding"
+        self.expr = ""
+
+        log.info(
+            f"{self.name} initialized with config:\nDatabase: {self.db_config}\nCase Config: {self.db_case_config}"
+        )
+
+        self._conn = None
+        self._cursor = None
+
+        try:
+            self._connect()
+            self._apply_system_settings()
+            if drop_old:
+                self._drop_table()
+                self._create_table()
+                self._create_index()
+        finally:
+            self._disconnect()
+
+    def _connect(self):
+        try:
+            self._conn = mysql.connect(
+                host=self.db_config["host"],
+                user=self.db_config["user"],
+                port=self.db_config["port"],
+                password=self.db_config["password"],
+                database=self.db_config["database"],
+            )
+            self._cursor = self._conn.cursor()
+        except mysql.Error:
+            log.exception("Failed to connect to SeekDB")
+            raise
+
+    def _disconnect(self):
+        if self._cursor:
+            self._cursor.close()
+            self._cursor = None
+        if self._conn:
+            self._conn.close()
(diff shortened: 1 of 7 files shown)


liuhao6741:
Hi, @XuanYang-cn  I  have fixed the lint error. can you approve the workflow and review the pr? thank you very much

liuhao6741:
> Hi, @XuanYang-cn I have fixed the lint error. can you approve the workflow and review the pr? thank you very much

sorry，I made a typo. I have fixed it. 


两项全部通过：

  make lint    ✅  black + ruff 全通过
  make unittest  ✅  1 passed（下载了约 6 分钟的数据集，之后会有缓存）

vectordb_bench/backend/clients/seekdb/seekdb.py
```diff
@@ -0,0 +1,285 @@
...
+        if p != m:
+            return p > m
+    return True
+
+
+class SeekDB(VectorDB):
+    supported_filter_types: list[FilterOp] = [
+        FilterOp.NonFilter,
+        FilterOp.NumGE,
+        FilterOp.StrEqual,
```
XuanYang-cn: SeekDB advertises `FilterOp.StrEqual`, but this backend never stores scalar labels: the table only has `id` and `embedding`, and `insert_embeddings()` ignores `labels_data`. Label-filter cases will be accepted by the assembler, but `prepare_filter()` later builds `WHERE id = 'label_...'`, which cannot return valid label-filter results. Please either add a label column and insert/search against `labels_data`, or remove `FilterOp.StrEqual` from `supported_filter_types`.

vectordb_bench/backend/clients/seekdb/seekdb.py
```diff
@@ -0,0 +1,285 @@
+import logging
+import re
+import struct
+from collections.abc import Generator
+from contextlib import contextmanager
+from typing import Any
+
+import mysql.connector as mysql
```
XuanYang-cn, line 8: This new backend requires `mysql-connector-python`, but the PR does not add a `seekdb` optional dependency in `pyproject.toml`. In a clean environment, importing `DB.SeekDB.init_cls` fails with `ModuleNotFoundError: No module named 'mysql'` unless another backend extra happened to install it. Please add a SeekDB extra, for example `seekdb = ["mysql-connector-python"]`.

XuanYang-cn (changes requested):
Requesting changes for the label-filter correctness issue and the missing optional dependency.

liuhao6741:
> Requesting changes for the label-filter correctness issue and the missing optional dependency.

Hi， @XuanYang-cn I have fixed the two issues. please review it again. Thanks very much.

XuanYang-cn (approved):

PR #772, opened 2026-05-01
Fix Skip File Validation When Offline with Existing Datasets

shaohuasong-fang:
## Description
When internet access is unavailable and a dataset has already been downloaded locally, the log output display it's still attempts file validation against remote servers, causing unnecessary failures.

## Finished display

Successfully bypassed Alibaba Cloud (OSS) and Amazon S3 file size validation checks when corresponding datasets are already available locally.

### DIFF
--- a/vectordb_bench/backend/data_source.py
+++ b/vectordb_bench/backend/data_source.py
@@ -86,9 +86,13 @@ class AliyunOSSReader(DatasetReader):
                 remote_file = pathlib.PurePosixPath("benchmark", dataset, file)
                 local_file = local_ds_root.joinpath(file)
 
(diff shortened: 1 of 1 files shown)


XuanYang-cn:
@shaohuasong-fang Thanks for the PR. I don’t think we should merge this approach.

  The current file-size validation is intentional. If a dataset was downloaded from the official remote source, we should keep validating that the local file still matches the remote metadata.

  Skipping validation whenever a local file exists can accept a corrupted or half-downloaded file, **for example if a previous download was interrupted.**

  I agree there is a valid offline/custom-dataset use case, but it should be explicit rather than changing the default behavior for official datasets.

shaohuasong-fang:
@XuanYang-cn 
Get it,thanks your explaining !

PR #773, opened 2026-05-09
(feat) Add Turbopuffer namespace pinning CLI

jamesgao-jpg:
## Summary
- Add `--pin-namespace/--no-pin-namespace` and `--pin-replicas` to the Turbopuffer benchmark CLI.
- PATCH Turbopuffer namespace metadata to request pinning before benchmark operations when explicitly enabled.
- Add `TurboPufferUnpin` CLI command to clear namespace pinning after benchmark runs.

## Test Plan
- `python -m py_compile vectordb_bench/backend/clients/turbopuffer/cli.py vectordb_bench/backend/clients/turbopuffer/config.py vectordb_bench/backend/clients/turbopuffer/turbopuffer.py vectordb_bench/cli/vectordbbench.py`
- Stubbed metadata PATCH smoke test for unpin payload (`{"pinning": null}`), no real Turbopuffer API calls made.

## Notes
- Full test suite was not run in the local checkout because the environment is missing the broader VectorDBBench dependency stack.

### DIFF
--- a/tests/test_turbopuffer_cli.py
+++ b/tests/test_turbopuffer_cli.py
@@ -12,8 +12,8 @@ def test_pin_namespace_is_applied_once_before_benchmark(monkeypatch):
         calls.append((method, payload, api_key, region, namespace, api_base_url))
         return {}
 
-    def fake_wait_for_pinning(api_key, region, namespace, replicas, api_base_url=None):
-        calls.append(("WAIT", replicas, api_key, region, namespace, api_base_url))
+    def fake_wait_for_pinning(api_key, region, namespace, replicas, api_base_url=None, timeout=None):
+        calls.append(("WAIT", replicas, api_key, region, namespace, api_base_url, timeout))
         return {"pinning": {"replicas": replicas, "status": {"ready_replicas": replicas}}}
 
     def fake_run(**kwargs):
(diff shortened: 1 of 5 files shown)


alexanderguzhva:
@jamesgao-jpg I'm not very familiar with VectorDBBench internals, just in case :) Nevertheless, this `TurboPufferUnpin` makes me confused, because (as a user) I don't understand why it needed to have it, and there's no description / anything that could explain me this.

jamesgao-jpg:
> @jamesgao-jpg I'm not very familiar with VectorDBBench internals, just in case :) Nevertheless, this `TurboPufferUnpin` makes me confused, because (as a user) I don't understand why it needed to have it, and there's no description / anything that could explain me this.

I've pondered about where to document this, at the end I can't think of a better location than where the code currently lies, since this is a rather niche thing with turbopuffer (they allow users to pin certain namespace for indefinite duration for more predictable performance). 
image

alexanderguzhva:
@jamesgao-jpg Please ensure that this is properly documented in readme.md on the https://github.com/zilliztech/VectorDBBench. Also, can this be implemented as a flag to plain Turbopuffer instead of a separate benchmark? I'm not saying whether this is needed to be implemented this way, but I'm asking to confirm the decision.

PR #775, opened 2026-05-11
feat: Add VectorDBBench Cloud Leaderboard benchmark cases and client support

jamesgao-jpg:
## Summary

This PR adds the VectorDBBench Cloud Leaderboard benchmark surface. The goal is to complement the existing raw-performance leaderboard with cloud-oriented cases that capture production behaviors managed vector database users care about: ingest readiness, payload-aware search, tenant-shaped traffic, cold latency, and cost-aware interpretation.

### New Cloud Leaderboard cases

- `CloudPayloadSearchCase`: measures search performance with explicit response payload profiles: `ids_only`, `scalar_label`, and `vector`. It supports unfiltered search, integer-filter search, and scalar-label filter search.
- `CloudInsertCase`: measures insert throughput and separates client insert completion from downstream readiness signals such as fully searchable and fully indexed.
- `CloudColdLatencyCase`: measures cold and warm serial latency so first-query and cache-sensitive serving behavior are visible instead of hidden by warm concurrent loops.
- `CloudMultiTenantSearchCase`: models SaaS-style tenant-routed workloads with deterministic tenant assignment and tenant-aware query routing.

### Runtime and metric plumbing

- Threads payload profile configuration through case config, runners, client search calls, metrics, and JSON result output.
- Adds payload byte estimation to result metrics so payload-heavy searches can be compared more explicitly.
- Adds a cold/warm search runner and task-runner integration for cold latency measurement.
- Adds concurrent insert readiness polling for cloud insert runs.
- Adds first-class result fields for cloud insert, cold latency, payload profile, and related cloud case metadata.

### Client support

- Milvus/Zilliz Cloud: adds payload output handling, scalar-label support, multitenant partition-key validation, and related schema checks.
- Pinecone: adds payload-profile search behavior, metadata/vector return handling, insert readiness polling through write/index LSNs where available, and namespace-based multitenant routing.
- turbopuffer: adds payload-profile search behavior, scalar payload label configuration, multitenant namespace support, write backpressure control, namespace pin/unpin CLI support for pinned benchmark runs, and explicit non-retryable partial tenant-write failure reporting.

### CLI, frontend, and docs

- Adds CLI options for payload profile, cloud filters, cold query count, insert batch size/duration, tenant settings, and turbopuffer pinning.
- Adds frontend case config entries for the cloud payload search cases.
- Adds a May 2026 VectorDBBench Cloud Leaderboard release note under `docs/release/2026-05-cloud-leaderboard.md`.
- Updates `README.md` to mention the new Cloud Leaderboard benchmark cases and link to the release note.

## Notes

- The raw Cloud Leaderboard result dump was intentionally removed from this PR. Result artifacts can be added later in a smaller, dedicated update or kept outside the source tree.
- Cloud cost and Pareto interpretation are described in the release note, but this PR focuses on benchmark case support and documentation rather than publishing a full result dataset.

## Test Plan

- `.venv/bin/python -m pytest tests/test_cloud_payload_search.py tests/test_cloud_payload_case.py tests/test_cloud_insert_case.py tests/test_cloud_cold_latency_case.py tests/test_multitenant_case.py tests/test_pinecone_multitenant.py tests/test_milvus.py -q -k 'not test_performance_1536d_50k'`
- `.venv/bin/python -m ruff check vectordb_bench/backend/clients/api.py vectordb_bench/backend/clients/turbopuffer/turbopuffer.py vectordb_bench/backend/runner/concurrent_runner.py vectordb_bench/cli/cli.py`


vectordb_bench/backend/cases.py
```diff
@@ -102,11 +109,21 @@ class Case(BaseModel):
...
+            k = config.K_DEFAULT
+        return self.payload_profile.estimated_bytes_per_query(k=k, dim=self.dataset.data.dim)
+
+    @property
+    def is_multitenant(self) -> bool:
+        return False
+
     @property
     def with_scalar_labels(self) -> bool:
         return self.filters.type == FilterOp.StrEqual
```
foxspy: This looks like a correctness issue: `with_scalar_labels` is currently derived only from `FilterOp.StrEqual`, but `PayloadProfile.SCALAR_LABEL` also requires scalar label data/schema even when there is no label filter.

For example, `CloudPayloadSearchCase(payload_profile="scalar_label")` uses `NonFilter`, so `with_scalar_labels` becomes false. The load path then does not create/load label data, while the search path still requests the label payload via `output_fields` / metadata / attributes. That means the benchmark may either fail at query time or measure a run where the requested scalar payload was never actually stored.

The multitenant path has a related issue: tenant routing labels and scalar payload/filter labels are treated as the same scalar-label concept in some insert paths. These should be separate fields/metadata concepts so a multitenant case can write both tenant routing information and the scalar payload label.

Suggested fix: make scalar-label materialization depend on both label filtering and payload profile, e.g. `filters.type == FilterOp.StrEqual or payload_profile == PayloadProfile.SCALAR_LABEL`, and keep tenant routing labels separate from scalar payload labels in the provider insert schemas.

vectordb_bench/cli/cli.py
```diff
@@ -184,14 +188,51 @@ def get_custom_case_config(parameters: dict) -> dict:
...
             "filter_rate": parameters["filter_rate"],
         }
     elif parameters["case_type"] == "LabelFilterPerformanceCase":
         custom_case_config = {
-            "dataset_with_size_type": parameters["dataset_with_size_type"],
+            "dataset_with_size_type": dataset_with_size_type,
             "label_percentage": parameters["label_percentage"],
         }
+    elif parameters["case_type"] == "CloudPayloadSearchCase":
+        custom_case_config = {
```
foxspy, line 212: This looks like a default/config propagation issue for the new cloud cases.

`CloudPayloadSearchCase` does not appear to receive `dataset_with_size_type` from the CLI custom case config, so a user-selected dataset can be silently ignored and the case falls back to its constructor default dataset. That would make the benchmark run against a different dataset than the CLI/config implies.

There is also an inconsistency in how missing defaults are handled across the cloud cases. `CloudMultiTenantSearchCase` receives `parameters["dataset_with_size_type"]` directly, which may be `None`, but the constructor later converts non-`DatasetWithSizeType` values with `DatasetWithSizeType(dataset_with_size_type)`. If `None` is passed through, the default CLI path can fail during case construction instead of using the case default.

Suggested fix: normalize optional CLI values before building `custom_case_config`. Either omit `dataset_with_size_type` when it is not explicitly provided, or pass a concrete default consistently. Also add CLI regression tests for omitted dataset/default behavior for `CloudPayloadSearchCase`, `CloudMultiTenantSearchCase`, and `CloudInsertCase`.

vectordb_bench/backend/clients/turbopuffer/turbopuffer.py
```diff
@@ -79,41 +215,112 @@ def insert_embeddings(
...
                         self._scalar_id_field: metadata,
-                        self._vector_field: embeddings,
+                        self._vector_field: vectors,
                     },
                     distance_metric=self.metric,
+                    disable_backpressure=self.db_case_config.disable_backpressure,
                 )
         except Exception as e:
             log.warning(f"Failed to insert. Error: {e}")
+            return 0, e
```
foxspy, line 302: This looks like a partial-failure accounting issue in the multitenant insert path.

When `tenant_labels_data` is present, this method writes one tenant namespace at a time and increments `inserted` after each successful tenant write. However, if a later tenant write fails, the outer `except` returns `0, e` even though earlier tenant namespaces may already have been written successfully.

That makes the caller's retry/statistics semantics misleading: the benchmark reports zero inserted rows for a batch that may have partially modified remote state, and a retry may re-write tenants that already succeeded.

Suggested fix: preserve the actual partial insert count on failure, or make partial tenant-write failures explicit/non-retryable with details about which tenants/counts were already written. A regression test where the first tenant write succeeds and the second tenant write fails would cover this behavior.

vectordb_bench/backend/cases.py
jamesgao-jpg: Fixed in `9343d01`.

Changes made:
- `with_scalar_labels` now includes `PayloadProfile.SCALAR_LABEL`, so scalar-label payload runs materialize label schema/data even without a label filter.
- `ConcurrentInsertRunner` now loads scalar labels when scalar-label materialization is required, not only for `FilterOp.StrEqual`.
- Milvus/Zilliz, Pinecone, and turbopuffer insert paths now keep tenant routing labels separate from scalar payload labels.
- Added regression coverage for unfiltered scalar-label payload loads and multitenant scalar payload inserts.
- Updated `docs/release/2026-05-cloud-leaderboard.md` to document the scalar-label and tenant-routing semantics.

Verification:
- `.venv/bin/python -m pytest tests/test_cloud_payload_search.py tests/test_cloud_insert_case.py tests/test_multitenant_case.py tests/test_pinecone_multitenant.py tests/test_milvus.py -q -k 'not test_performance_1536d_50k'` -> `55 passed, 1 deselected`
- `.venv/bin/python -m ruff check` on touched production files -> `All checks passed`

vectordb_bench/cli/cli.py
jamesgao-jpg: Fixed in `552155c`.

I verified the issue was real on the PR branch:
- `CloudPayloadSearchCase` dropped an explicit CLI `dataset_with_size_type`, so the case fell back to LAION 100M.
- `CloudMultiTenantSearchCase` passed `dataset_with_size_type=None`, which failed construction instead of using the case default.

Changes made:
- `CloudPayloadSearchCase` now propagates an explicitly selected dataset and omits the field when absent, preserving its LAION 100M constructor default.
- `CloudMultiTenantSearchCase` now omits absent dataset config, preserving its `CohereLarge` constructor default instead of passing `None`.
- `CloudInsertCase` keeps the existing CLI default of Medium Cohere when omitted.
- Added CLI regression coverage for payload explicit/omitted dataset behavior, multitenant omitted dataset behavior, and insert omitted/default behavior.

Verification:
- `.venv/bin/python -m pytest tests/test_cloud_payload_search.py tests/test_cloud_payload_case.py tests/test_cloud_insert_case.py tests/test_cloud_cold_latency_case.py tests/test_multitenant_case.py tests/test_pinecone_multitenant.py tests/test_milvus.py -q -k 'not test_performance_1536d_50k'` -> 82 passed, 1 deselected
- `.venv/bin/python -m ruff check vectordb_bench/cli/cli.py` -> passed

vectordb_bench/backend/clients/turbopuffer/turbopuffer.py
jamesgao-jpg: Fixed in `c1164ce`.

I verified the concern was real: TurboPuffer's multitenant insert path could successfully write one tenant namespace, then return `(0, error)` if a later tenant write failed. The concurrent insert runner also retried any returned error, so this could re-write tenants that had already succeeded.

Changes made:
- Added explicit non-retryable partial-insert error metadata: `inserted_count`, `successful_tenants`, `failed_tenant`, and `failed_tenant_count`.
- TurboPuffer multitenant insert now returns the actual partial count and that detailed non-retryable error when a tenant write fails after prior tenant writes.
- `ConcurrentInsertRunner` now stops immediately on non-retryable insert errors instead of retrying the whole batch.
- Added regression coverage for first-tenant-success/second-tenant-failure and for runner non-retry behavior.

Verification:
- `.venv/bin/python -m pytest tests/test_cloud_payload_search.py tests/test_cloud_payload_case.py tests/test_cloud_insert_case.py tests/test_cloud_cold_latency_case.py tests/test_multitenant_case.py tests/test_pinecone_multitenant.py tests/test_milvus.py -q -k 'not test_performance_1536d_50k'` -> 84 passed, 1 deselected
- `.venv/bin/python -m ruff check vectordb_bench/backend/clients/api.py vectordb_bench/backend/clients/turbopuffer/turbopuffer.py vectordb_bench/backend/runner/concurrent_runner.py` -> passed

vectordb_bench/backend/runner/concurrent_runner.py
```diff
@@ -158,13 +183,18 @@ def _next_batch(self) -> tuple[list[list[float]], list[int], list[str] | None] |
         del emb_np
 
         labels_data = None
-        if self.filters.type == FilterOp.StrEqual:
+        if self.filters.type == FilterOp.StrEqual or self.with_scalar_labels:
+            label_field = self.filters.label_field if self.filters.type == FilterOp.StrEqual else "labels"
             if self.dataset.data.scalar_labels_file_separated:
-                labels_data = self.dataset.scalar_labels[self.filters.label_field][all_metadata].to_list()
+                labels_data = self.dataset.scalar_labels[label_field][all_metadata].to_list()
```
XuanYang-cn, line 196: Scalar-label payload runs still fail for datasets whose labels live in `scalar_labels.parquet`.

`CloudPayloadSearchCase(payload_profile="scalar_label")` sets `with_scalar_labels=True`, but `DatasetManager.prepare()` only hydrates `dataset.scalar_labels` when the active filter is `StrEqual`. For the default unfiltered scalar-label payload case, this line dereferences `None` for separated-label datasets like LAION/Cohere.

Please load separated scalar labels during dataset preparation whenever the case requires scalar-label materialization, not only when the active filter is `StrEqual`, and add a regression test for the separated-label path.

vectordb_bench/backend/clients/pinecone/pinecone.py
```diff
@@ -87,7 +227,29 @@ def insert_embeddings(
...
+                        self._record_write_lsn(write_lsn)
+                else:
+                    batch_tenant_labels = tenant_labels_data[batch_start_offset:batch_end_offset]
+                    for tenant in sorted(set(batch_tenant_labels)):
+                        tenant_insert_datas = [
+                            insert_data
+                            for insert_data, tenant_label in zip(insert_datas, batch_tenant_labels, strict=True)
+                            if tenant_label == tenant
+                        ]
+                        self._multitenant_insert_counts[tenant] = self._multitenant_insert_counts.get(tenant, 0) + len(
```
XuanYang-cn: Pinecone multitenant inserts still account partial tenant writes incorrectly.

This increments `_multitenant_insert_counts` before the tenant namespace upsert succeeds, while `insert_count` is advanced only after all tenant writes in the batch complete. If tenant A succeeds and tenant B fails, the method returns `(0, error)` even though tenant A was written, and `_multitenant_insert_counts` includes the failed tenant too.

Please mirror the TurboPuffer partial-insert handling here: update tenant counts only after successful writes and return a non-retryable partial failure with the actual inserted count and tenant details.

vectordb_bench/backend/task_runner.py
```diff
@@ -245,17 +282,104 @@ def _run_streaming_case(self) -> Metric:
...
+            batch_size=self.ca.batch_size,
+            duration=self.ca.duration,
+            **runner_kwargs,
+        )
+        count = runner.task()
+        insert_done = time.perf_counter()
+        with self.db.init():
+            status = self.db.poll_insert_readiness(count)
+            searchable_started = time.perf_counter()
+            while not status["fully_searchable"]:
```
XuanYang-cn, line 371: Cloud insert readiness polling can hang forever here.

Both readiness loops poll without any deadline. A service-side indexing stall, an API/status mismatch, or incorrect expected-count state will leave the benchmark process stuck indefinitely after the insert phase has completed.

Please add a readiness timeout/deadline, preferably configurable from the cloud insert case, and include the last readiness status in the raised error so failed runs are diagnosable.

vectordb_bench/backend/clients/turbopuffer/cli.py
```diff
@@ -10,54 +10,191 @@
...
+        )
+
+
 @cli.command()
 @click_parameter_decorators_from_typed_dict(TurboPufferIndexTypedDict)
 def TurboPuffer(**parameters: Unpack[TurboPufferIndexTypedDict]):
     from .config import TurboPufferConfig, TurboPufferIndexConfig
 
+    pin_target_namespace_count = len(target_namespaces_for_pinning(parameters)) if parameters["pin_namespace"] else 0
+    if parameters["pin_namespace"]:
```
XuanYang-cn: `--dry-run --pin-namespace` still mutates remote TurboPuffer state.

This calls `pin_namespaces_once()` before `run()` sees `dry_run`, and `pin_namespaces_once()` PATCHes namespace metadata and waits for pinning. Since pinning reserves persistent service resources, a dry run can still create billing/state side effects.

Please skip pinning when `parameters["dry_run"]` is true, or move the pinning action behind the same dry-run guard used for benchmark execution. A CLI test should assert that dry-run with `--pin-namespace` does not call the metadata PATCH helper.

XuanYang-cn (changes requested):
Requesting changes. I left the blocking findings inline.

vectordb_bench/models.py
```diff
@@ -276,6 +278,68 @@ class TestResult(BaseModel):
...
+                "case_id": case_config.case_id.value,
+                "custom_case": case_config.custom_case,
+            }
+
+        return case_config.model_dump(mode="json")
+
+    def model_dump_for_output(self) -> dict:
+        output = self.model_dump(mode="json", serialize_as_any=True)
+        for idx, case_result in enumerate(self.results):
+            output["results"][idx]["metrics"] = self._output_metrics_for_case(case_result)
```
XuanYang-cn, line 340: This projection makes the new cloud result files unreadable by the existing result collector. `ResultCollector.collect()` calls `TestResult.read_file(..., trans_unit=True)` for every result file, and that conversion still indexes `max_load_count`, `serial_latency_p99`, and related legacy metric keys. Since the CloudInsert and CloudColdLatency branches now write only case-specific metrics, collecting a cloud insert result raises `KeyError: 'max_load_count'` before model validation fills defaults. Please either keep the legacy metric keys/defaults in the serialized output or make the `trans_unit` conversion case-aware, and add a regression that reads a cloud insert/cold-latency result through `ResultCollector.collect()`.

XuanYang-cn (changes requested):
I found one additional required issue in the result read/write path.

These existing unresolved review threads still apply on the current head and should remain blockers: scalar-label payload loads with separated `scalar_labels.parquet`, Pinecone multitenant partial-insert accounting, cloud insert readiness polling without a timeout, and TurboPuffer `--dry-run --pin-namespace` mutating remote state before dry-run handling.

vectordb_bench/backend/runner/concurrent_runner.py
jamesgao-jpg: Fixed in `d1b545b`. `DatasetManager.prepare()` now accepts a `with_scalar_labels` signal and hydrates separated `scalar_labels.parquet` when either a `StrEqual` label filter or scalar-label payload materialization requires it. `CaseRunner._pre_run()` passes `self.ca.with_scalar_labels` into dataset preparation, so the default unfiltered `CloudPayloadSearchCase(payload_profile="scalar_label")` path loads `dataset.scalar_labels` before `ConcurrentInsertRunner` reads it.

Added regression coverage for both pieces:
- `test_dataset_prepare_loads_separated_scalar_labels_for_scalar_payload`
- `test_pre_run_prepares_separated_scalar_labels_for_scalar_payload`

Verification: `python3.11 -m pytest tests/test_cloud_payload_search.py tests/test_cloud_insert_case.py -q` -> `40 passed`.

vectordb_bench/backend/clients/pinecone/pinecone.py
jamesgao-jpg: Fixed in `7d2a4d6`. Pinecone multitenant inserts now update `_multitenant_insert_counts` only after a tenant namespace upsert succeeds. The insert path tracks actual inserted rows and successful tenants; if a later tenant upsert fails, it returns a non-retryable `PartialInsertError` with `inserted_count`, `successful_tenants`, `failed_tenant`, and `failed_tenant_count`, matching the TurboPuffer handling. This prevents the concurrent insert runner from retrying a partially written tenant batch and prevents readiness expectations from including the failed tenant write.

Added regression coverage in `test_pinecone_multitenant_partial_insert_failure_is_explicit`.

Verification: `python3.11 -m pytest tests/test_pinecone_multitenant.py tests/test_cloud_insert_case.py::test_concurrent_insert_runner_does_not_retry_non_retryable_insert_errors -q` -> `4 passed`.

vectordb_bench/backend/task_runner.py
jamesgao-jpg: Fixed in `589b854`. Cloud insert readiness polling now supports an explicit timeout/deadline and includes the last readiness status in the raised `TimeoutError` for diagnosis. The timeout is intentionally opt-in: `CloudInsertCase.readiness_timeout` defaults to `None`, preserving the previous behavior of waiting indefinitely unless the user sets `--cloud-insert-readiness-timeout` or `CLOUD_INSERT_READINESS_TIMEOUT`. The poll interval is also configurable via `--cloud-insert-readiness-poll-interval` / `CLOUD_INSERT_READINESS_POLL_INTERVAL` and defaults to the prior 5s interval.

Added regression coverage for:
- default `CloudInsertCase` has no readiness timeout
- CLI propagation of explicit timeout/poll interval
- stalled readiness raises `TimeoutError` with `last_status` when a timeout is explicitly configured

Verification: `python3.11 -m pytest tests/test_cloud_insert_case.py -q` -> `34 passed`.

vectordb_bench/backend/clients/turbopuffer/cli.py
jamesgao-jpg: Fixed in `fde0b32`. TurboPuffer CLI now skips `pin_namespaces_once()` when `--dry-run` is true, so `--dry-run --pin-namespace` no longer PATCHes namespace metadata or waits for pinning. The config still records `pin_namespace_requested=True` and target count for visibility. Added dry-run regression coverage for single and multitenant pinning. Verification: `python3.11 -m pytest tests/test_turbopuffer_cli.py -q` -> `4 passed`.

vectordb_bench/models.py
jamesgao-jpg: Fixed in `c78c979`. `TestResult.read_file(..., trans_unit=True)` now converts display units only for metric keys that are present, so compact `CloudInsertCase` and `CloudColdLatencyCase` JSON files can be loaded through `ResultCollector.collect()` without requiring legacy `max_load_count` or `serial_latency_p99` fields. The old P95 default is still preserved for legacy search-style files that have P99 but lack P95. Added collector regression coverage for both compact cloud result formats. Verification: `python3.11 -m pytest tests/test_cloud_insert_case.py tests/test_cloud_cold_latency_case.py -q` -> `53 passed`.

vectordb_bench/backend/cases.py
```diff
@@ -102,14 +109,24 @@ class Case(BaseModel):
...
+        return self.payload_profile.estimated_bytes_per_query(k=k, dim=self.dataset.data.dim)
+
+    @property
+    def is_multitenant(self) -> bool:
+        return False
+
     @property
     def with_scalar_labels(self) -> bool:
-        return self.filters.type == FilterOp.StrEqual
+        return self.filters.type == FilterOp.StrEqual or self.payload_profile == PayloadProfile.SCALAR_LABEL
```
XuanYang-cn, line 129: This makes scalar-label payloads require scalar-label schema/data, but the runner reuse key still treats an `ids_only`/`vector` performance case and a `scalar_label` case on the same DB/config/dataset as the same loaded collection. In `_async_task_v2`, the later case then runs with `drop_old=False`; if the first case loaded without scalar labels, the collection lacks the scalar-label field/data while this case still requests scalar-label payloads. I reproduced the reuse condition locally with two `CloudPayloadSearchCase` instances on the same dataset: `with_scalar_labels` is `False` vs `True`, but the two `CaseRunner`s compare equal. Please include the scalar-label/schema requirement, or payload profile, in the reuse key, or otherwise ensure scalar-label-required cases only reuse collections that were loaded with scalar labels.

vectordb_bench/backend/clients/turbopuffer/turbopuffer.py
```diff
@@ -59,12 +139,68 @@ def _create_client(self) -> tpuf.Turbopuffer:
...
+                "TurboPuffer pinning requested for %s: replicas=%s ready_replicas=%s",
+                namespace,
+                pinning.get("replicas", self.pin_replicas) if isinstance(pinning, dict) else self.pin_replicas,
+                status.get("ready_replicas"),
+            )
+        self._pinning_applied = True
+
+    def _target_namespaces_for_pinning(self) -> list[str]:
+        if self.multitenant_tenant_labels:
+            return [self._namespace_name_for_tenant(tenant) for tenant in self.multitenant_tenant_labels]
```
XuanYang-cn, line 177: The multitenant namespace list is only used for pinning. `optimize()` still calls `self.ns.hint_cache_warm()` on the base namespace created in `init()`, while multitenant inserts and searches go through `_namespace_for_tenant(tenant)`. A `CloudMultiTenantSearchCase` therefore warms `namespace`, often empty, and sleeps, leaving the actual tenant namespaces cold before the performance search. Please warm each tenant namespace, or the touched tenant set, when `multitenant_tenant_labels` is set, and add a regression that `hint_cache_warm()` is called on `mt_tenant_*` rather than the base namespace.

XuanYang-cn (changes requested):
I found two blocking issues that can still produce invalid Cloud Leaderboard measurements: scalar-label cases can reuse a collection loaded without scalar labels, and TurboPuffer multitenant optimize warms the base namespace instead of the tenant namespaces.

vectordb_bench/backend/cases.py
jamesgao-jpg: Fixed in `a949875`.

The collection reuse key now includes the schema/data requirement via `ca.with_scalar_labels`, so a scalar-label payload case will not reuse a collection that was loaded by an ids-only/vector case without scalar label fields. I used `with_scalar_labels` rather than the raw `payload_profile` so compatible ids-only/vector runs can still reuse the same loaded collection, while scalar-label-required runs get a separate load path.

Added a regression in `tests/test_cloud_payload_case.py` that constructs ids-only and scalar-label `CloudPayloadSearchCase` runners on the same DB/config/dataset and verifies they no longer compare equal.

vectordb_bench/backend/clients/turbopuffer/turbopuffer.py
jamesgao-jpg: Fixed in `a949875`.

The base namespace warmup was wrong for multitenant TurboPuffer because inserts/searches use tenant namespaces. I changed multitenant optimize behavior so it no longer warms the unused base namespace by default.

I also made tenant warmup explicit with a TurboPuffer-only `--multitenant-warmup-policy` option:

- `none` (default): skip tenant namespace cache warmup for multitenant runs, which better represents small/inconsistent tenant traffic and avoids silently biasing latency.
- `all`: warm every configured tenant namespace when the benchmark intentionally wants proactively warmed tenant namespaces.

Added regressions that default multitenant optimize does not warm `base`, and `all` warms `mt_tenant_*` namespaces. The release doc now calls out the policy and when to use it.

vectordb_bench/backend/task_runner.py
```diff
@@ -64,6 +67,7 @@ def __eq__(self, obj: any):
                 and self.config.db == obj.config.db
                 and self.config.db_case_config == obj.config.db_case_config
                 and self.ca.dataset == obj.ca.dataset
+                and self.ca.with_scalar_labels == obj.ca.with_scalar_labels
```
XuanYang-cn: must-change: the runner reuse key still ignores multitenant routing parameters. Adjacent `CloudMultiTenantSearchCase`s on the same DB/config/dataset but different `tenant_count`, `tenant_prefix`, or `tenant_id_width` compare equal, so `_async_task_v2` sets `drop_old=False` and the later case searches tenant labels or namespaces that were never loaded. Add the multitenant routing key, for example the full `tenant_labels()` or the tenant config tuple, to `__eq__` and `__hash__`.

vectordb_bench/backend/runner/concurrent_runner.py
```diff
@@ -109,20 +117,34 @@ def _insert_batch_with_retry(
...
+            "embeddings": embeddings,
+            "metadata": metadata,
+            "labels_data": labels_data,
+        }
+        if tenant_labels_data is not None:
+            insert_kwargs["tenant_labels_data"] = tenant_labels_data
+        insert_count, error = db.insert_embeddings(**insert_kwargs)
         if error is not None:
             log.warning(f"Insert failed, try_idx={retry_idx}, Exception: {error}")
+            if getattr(error, "non_retryable", False):
```
XuanYang-cn, line 135: must-change: non-retryable partial insert errors only stop the current worker. The other submitted worker loops keep draining the shared dataset until `wait_all()` returns, so after one tenant batch reports a partial remote write, this runner can keep writing more batches and make the failure state larger than the reported partial count. Add a shared stop flag or cancellation path that prevents `_next_batch()` from handing out more batches after a non-retryable insert error.

vectordb_bench/backend/clients/pinecone/pinecone.py
```diff
@@ -87,7 +228,48 @@ def insert_embeddings(
...
+                                failed_tenant=tenant,
+                                failed_tenant_count=len(tenant_insert_datas),
+                                cause=e,
+                            )
+                        write_lsn = self._extract_lsn(upsert_res, "x-pinecone-request-lsn")
+                        if write_lsn is not None:
+                            self._record_write_lsn(write_lsn)
+                        insert_count += len(tenant_insert_datas)
+                        successful_tenants[tenant] = successful_tenants.get(tenant, 0) + len(tenant_insert_datas)
+                        self._multitenant_insert_counts[tenant] = self._multitenant_insert_counts.get(tenant, 0) + len(
```
XuanYang-cn: must-change: `_multitenant_insert_counts` is updated by all insert worker threads without synchronization. The read-modify-write can lose successful tenant rows under `load_concurrency > 1`, and `poll_insert_readiness()` then waits for too few vectors and can mark the run searchable/indexed early. Protect this map with a lock or return per-worker tenant counts and merge them after workers finish.

XuanYang-cn (changes requested):
Reviewed head a949875. I found blocking issues in the cloud benchmark state handling: multitenant case reuse can run against the wrong tenant layout, non-retryable partial insert failures do not stop the remaining insert workers, and Pinecone multitenant readiness counts are racy under concurrent insert.

vectordb_bench/backend/task_runner.py
jamesgao-jpg: Fixed in `73354cb`. The runner reuse logic now uses an explicit loaded-data reuse key instead of the previous partial equality tuple. That key includes the DB physical target config, DB case config, Doris' case-derived collection name, dataset identity, scalar-label schema requirement, multitenant layout, and the multitenant routing tuple (`tenant_count`, `tenant_prefix`, `tenant_id_width`, `tenant_distribution`).

I also changed `_async_task_v2` to track the latest successful loaded reuse key instead of the latest successful runner, so a search-only case cannot suppress a later full load. Added focused regression coverage for different multitenant routing, single-tenant vs multitenant layout, safe ids/vector payload reuse, different physical DB targets, Doris table naming, and search-only followed by full-load behavior.

Verified with `make lint`, `git diff --check`, and the focused cloud/reuse test set (`94 passed`).

vectordb_bench/backend/runner/concurrent_runner.py
jamesgao-jpg: Fixed in `6b64cd8`.

The runner now has a shared `_stop_event`: when any worker raises from insert processing, `_worker_loop()` sets the event and re-raises; `_next_batch()` checks the event both before and after taking the iterator lock, so other workers stop receiving new dataset batches once the run has observed a fatal insert failure.

I kept this at the `ConcurrentInsertRunner` level rather than making it tenant-specific. `PartialInsertError` is the explicit non-retryable case we added for namespace-per-tenant multitenant inserts, but retry exhaustion for ordinary insert errors also means the run is already failing. In both cases, handing out more batches only enlarges the mutated remote state. Successful inserts and transient errors that recover within retries keep the same behavior.

Added regression coverage for both paths: non-retryable partial insert errors and retry-exhausted regular insert errors. Verified with the cloud/multitenant PR test subset (`96 passed`), `make lint`, and `git diff --check`.

vectordb_bench/backend/clients/pinecone/pinecone.py
jamesgao-jpg: Fixed in `a83a85a`.

I added a Pinecone-specific lock around `_multitenant_insert_counts` updates and snapshot reads, plus a regression test that forces two concurrent insert workers to update the same tenant count. Without the lock the test records one successful row after two successful upserts; with the lock it records both.

One caveat from checking the case flow: `poll_insert_readiness()` is currently called by `CloudInsertCase`, while the named `CloudMultiTenantSearchCase` follows the performance load path. So the early-readiness consequence is mainly relevant if Pinecone multitenant insert is combined with cloud-insert readiness or reused directly. The underlying shared mutable state issue still exists under `load_concurrency > 1`, so the fix keeps the bookkeeping correct without changing tenant routing or insert behavior.

I documented the provider-sensitive part inline: Pinecone tracks per-tenant counts because its readiness path is count-based per namespace; providers such as TurboPuffer only track touched tenant membership, so they do not have the same lost-increment issue.

Verified with the cloud/multitenant PR subset (`97 passed`), `make lint`, and `git diff --check`.

XuanYang-cn (approved):

PR #785, opened 2026-05-20
Add support for Garnet

### LINKED ISSUE #1176: Support filtered workloads in VectorDBBench
Add support for the filtered workloads in the Garnet driver for VectorDBBench

metajack:
Garnet is a new remote cache-store from Microsoft Research which is compatible with Redis. It now has vector set support, and this PR adds support for that to VectorDBBench.

Note that while this Garnet driver will work unmodified for Redis with vector sets, it will eventually get more features that are unique to Garnet, since Garnet plans to support a wider array of vector formats and quantizers than Redis currently has.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -41,6 +41,7 @@ dependencies = [
     "pymilvus<3.0.0", # with pandas, numpy
     "hdrhistogram>=0.10.1",
     "ujson",
+    "PyYAML",
 ]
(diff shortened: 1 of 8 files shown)


metajack:
This now has support for the filtering cases.

XuanYang-cn:
@metajack Please fix the conflict

PR #791, opened 2026-05-31
feat: Add YDB support to the benchmark

### LINKED ISSUE #792: Add support for YDB as a benchmark target database
[YDB](https://ydb.tech) is a distributed SQL database which supports [vector search](https://ydb.tech/docs/en/concepts/query_execution/vector_search?version=main), including the exact and indexed (approximate) search.

In order to provide a basic comparison of YDB's capabilities to other systems, we would like to propose the code supporting YDB as the VectorDBBench target.

zinal:
[YDB](https://ydb.tech) is a distributed SQL database which supports [vector search](https://ydb.tech/docs/en/concepts/query_execution/vector_search?version=main), including the exact and indexed (approximate) search.

This PR adds basic support for YDB to the VectorDBBench code.

Fixes #792

## Notes for maintainers

Adding a YDB client requires small, opt-in extensions to shared benchmark infrastructure. The goal is to support gRPC-based clients and case-specific index DDL without changing behavior for existing databases.

All changes follow the same pattern already used for `thread_safe` in `concurrent_runner.py:` boolean capability flags on `VectorDB`, checked via `getattr(..., False)`. No database-specific `if db == DB.X` branches remain in shared runners.

YDB uses a gRPC driver and builds vector_kmeans_tree indexes after data load. Index DDL depends on the case filter type. That creates three practical constraints:

1. Subprocess + pickle — the YDB driver/session pool cannot be pickled when `ProcessPoolExecutor` forks or spawns a child process.
2. Filters at init time — index shape must be known when `optimize()` runs, not only during search.
3. Per-case table names — same requirement as for Doris: avoid cross-case interference when running multiple benchmarks against one cluster.

These are client-specific needs, not general benchmark changes.


XuanYang-cn:
I found one blocking issue in the YDB filtered-index path:

`vectordb_bench/backend/clients/ydb/ydb_client.py` builds filtered vector indexes as `ON (id, embedding)` or `ON (labels, embedding)`, but `_index_probe_query()` probes `VIEW vector_idx` without the matching `WHERE` predicate. For filtered YDB vector indexes, queries through the index need the filter columns constrained, so the readiness probe can fail even after the index is usable. `_probe_index_status()` then treats that as still building, and `_wait_for_index()` can sit until the 7200s timeout on filtered cases.

Please update the readiness probe to use the same filter predicate as the search path, or use a schema/index-status check that does not depend on an unfiltered `VIEW` query.

Also, this PR is currently conflicting with `main`. Please rebase/merge `main` and fix the conflicts.

PR #793, opened 2026-06-05
feat: add VolcanoMilvus DISKANN support

lyxiong0:
## Summary

  Add VolcanoMilvus as a new VectorDBBench target by extending the existing Milvus client with Volcano-specific DISKANN behavior.

  ## What's added

  - New VolcanoMilvus client implementation under vectordb_bench/backend/clients/volcano_milvus/
  - New CLI command: vectordbbench volcanomilvusdiskann ...
  - Volcano DISKANN config support:
      - build params: max_degree, build_search_list, legacy, store_strategy, quant_type, num_threads, distance_strategy

  - Load behavior knobs: load_reqs_size and load_after_compaction
  - Frontend integration for Load, Performance, and Streaming benchmark cases
  - README/model enum updates so VolcanoMilvus is discoverable as a supported client
  - Unit tests for CLI parsing, frontend config exposure, load/search params, and DISKANN-only case mapping

  ## Supported index
 Only DISKANN is exposed for VolcanoMilvus in this change. Other index types do not fall back to the upstream Milvus configs silently.

  ## Example
```bash
vectordbbench volcanomilvusdiskann \
  --case-type Performance768D10M \
  --k 100 \
  --num-concurrency 400 \
  --uri <uri> \
  --user-name <user-name> \
  --password <pass> \
  --search-list 300 \
  --max-degree 48  \
  --replica-number 1 \
  --num-shards 1 \
  --load-after-compaction  \
  --distance-strategy "quant then full" \
  --num-threads 64 \
  --concurrency-duration 30 \
  --enable-thp 
  --enable-prefetch
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -36,7 +36,7 @@ All the database client supported
 
 | Optional database client | install command                             |
 |--------------------------|---------------------------------------------|
-| pymilvus, zilliz_cloud (*default*)     | `pip install vectordb-bench`                |
+| pymilvus, zilliz_cloud, volcanomilvus (*default*)     | `pip install vectordb-bench`                |
 | qdrant                   | `pip install vectordb-bench[qdrant]`        |
 | pinecone                 | `pip install vectordb-bench[pinecone]`      |
 | weaviate                 | `pip install vectordb-bench[weaviate]`      |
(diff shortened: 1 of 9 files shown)


xiaofan-luan:
unfortunately, there is only one milvus, which is opensourced milvus.
We are happy to take contribution and update opensource milvus benchmark. Unfortunatelly any other opensource milvus branch won't be listed.

PR #800, opened 2026-06-17
Update Hologres client to use binary dumper

zpatronus:
Reduce Hologres client-side overhead.

Changes:
- Add a custom binary dumper (`HoloFloat4Array`) for float4[] serialization.

### DIFF
--- a/vectordb_bench/backend/clients/hologres/hologres.py
+++ b/vectordb_bench/backend/clients/hologres/hologres.py
@@ -2,6 +2,7 @@
 
 import json
 import logging
+import struct
 from collections.abc import Generator
 from contextlib import contextmanager
 from io import StringIO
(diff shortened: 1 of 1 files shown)


vectordb_bench/backend/clients/hologres/hologres.py
```diff
@@ -2,20 +2,55 @@
...
+    oid = 1021  # PostgreSQL OID for float4[] (_float4)
+
+    def dump(self, obj: HoloFloat4Array) -> bytes:
+        # PostgreSQL binary array format:
+        # Header: ndim(4) + has_null(4) + OID elemtype(4) + dim_size(4) + lower_bound(4) = 20 bytes
+        # For each element: itemlen(4) + float32(4) = 8 bytes per element
+        header = struct.pack(">iiIii", 1, 0, 700, len(obj.data), 1)
+        elements = b""
+        for x in obj.data:
+            elements += struct.pack(">if", 4, x)  # 4 = length of float32, x = the float value
```
XuanYang-cn: nit: line 44 rebuilds the accumulated bytes object on every element because bytes is
  immutable. This is in the search serialization hot path, and a quick local benchmark
  shows the cost grows with vector dimension. Replace the loop with a joined sequence or
  preallocated buffer, e.g. return header + b"".join(struct.pack(">if", 4, x) for x in
  obj.data).

zpatronus: Fixed, thanks!

PR #801, opened 2026-06-18
fix(runner): add 10s cooldown between concurrent and serial search ph…

zpatronus:
…ases

The serial search phase measures single-query latency under idle conditions. Previously it launched immediately after the concurrent phase, while the end-to-end path may still be in a saturated state. This significantly affects the accuracy of serial latency, especially p99/p95, because the first few queries experience an environment similar to the concurrent test rather than the intended single-user condition.

---

This improves serial latency accuracy for **all database backends**.

### DIFF
--- a/vectordb_bench/backend/task_runner.py
+++ b/vectordb_bench/backend/task_runner.py
@@ -318,6 +318,12 @@ class CaseRunner(BaseModel):
                         m.conc_latency_avg_list,
                     ) = search_results
                 if TaskStage.SEARCH_SERIAL in self.config.stages:
+                    if TaskStage.SEARCH_CONCURRENT in self.config.stages:
+                        cooldown = 10
+                        log.info(
+                            f"Cooldown {cooldown}s before serial search to ensure a stable measurement environment"
+                        )
+                        time.sleep(cooldown)
                     search_results = self._serial_search()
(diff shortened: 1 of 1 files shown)


XuanYang-cn:
@zpatronus I'd suggest closing this. The PR assumes serial latency should reflect a **single-user / idle** condition, and a cooldown brings it closer to that.

But serial isn't measuring idle — it measures **warm best-case latency**. That's why serial runs after concurrent on purpose: the concurrent phase replays the *same* queries to warm the server, and serial then measures that warm state. So post-concurrent isn't contamination to recover from — it's exactly the intended condition. A cooldown only pushes the server back toward cold, away from the target.

Also, `10s` is an arbitrary constant — nothing verifies the system settled, and one value can't fit all backends.

zpatronus:
@XuanYang-cn  Fair point on the warm state. The cooldown isn't meant to evict data caches or similar thing, just to drain transient things like in-flight requests, network buffers, etc. 
Agreed that 10s is arbitrary. Even 1s would be sufficient for what I'm describing. Alternatively, would a configurable --serial-cooldown-ms parameter (default 0) work? That way backends that need it can opt in. Happy to implement either.

zpatronus:
#803 implements the configurable parameter.

PR #802, opened 2026-06-18
Add OpenSearch Serverless (AOSS) support

norrishuang:
## Summary

Add support for Amazon OpenSearch Serverless (AOSS) to the awsopensearch client.

## Changes

- **config.py**: Add `is_serverless` and `aws_region` fields; use AWS SigV4 authentication via `requests-aws4auth` for AOSS connections
- **cli.py**: Add `--serverless` flag and `--aws-region` option
- **aws_opensearch.py**: Adapt client for AOSS constraints:
  - Skip unsupported operations (cluster settings, force merge, manual refresh, replica updates, warmup API)
  - Use smaller batch size (100) for bulk inserts
  - Store id as a document field (AOSS doesn't support custom `_id`)
  - Retrieve id from `_source` in search results
  - Remove `engine` and `encoder` from method config (AOSS manages internally)
  - Disable `http_compress` to avoid SigV4 checksum verification failures
- **README.md**: Add OpenSearch Serverless section with usage example and notes

## Usage

```shell
vectordbbench awsopensearch --db-label aoss \
  --serverless --aws-region us-east-1 \
  --host <collection-id>.aoss.us-east-1.on.aws --port 443 \
  --case-type Performance768D1M \
  --m 16 --ef-construction 200 --ef-search 40 \
  --number-of-shards 8 --number-of-replicas 0 \
  --engine faiss --metric-type cosine \
  --num-concurrency 80,100,120
```

## Prerequisites

- AWS credentials configured
- `requests-aws4auth` installed
- IAM identity policy with `aoss:APIAccessAll`
- AOSS Data Access Policy granting index/collection permissions

## Testing

Tested against a live AOSS collection with 1M 768-dim vectors (Cohere dataset). Data loading and index creation verified successfully.


XuanYang-cn (commented):
I left a few inline comments for the AOSS paths that look like they still need fixes before merge.

vectordb_bench/backend/clients/aws_opensearch/aws_opensearch.py
```diff
@@ -222,25 +243,47 @@ def _insert_with_single_client(
...
+                    index_data = {"index": {"_index": self.index_name}}
+                else:
+                    index_data = {"index": {"_index": self.index_name, self.id_col_name: batch_metadata[j]}}
+
+                if self.with_scalar_labels and self.case_config.use_routing and batch_labels is not None:
+                    index_data["routing"] = batch_labels[j]
+                insert_data.append(index_data)
+
+                other_data = {self.vector_col_name: batch_embeddings[j]}
+                if self._is_serverless:
```
XuanYang-cn, line 279: This serverless ID handling only exists in the single-client insert path. When `number_of_indexing_clients > 1`, `insert_embeddings()` dispatches to `_insert_with_multiple_clients()`, which still sends `self.id_col_name` as the bulk action `_id` and never stores the benchmark id in `_source`. Since the PR notes AOSS does not support custom `_id`, serverless loads with multiple indexing clients will still fail. Please apply the same no-custom-`_id` / `_source.id` / small-batch behavior to the multi-client path too, or force serverless through the single-client path.

vectordb_bench/backend/clients/aws_opensearch/aws_opensearch.py
```diff
@@ -402,24 +445,37 @@ def search_embedding(
...
+                resp = self.client.search(
+                    index=self.index_name,
+                    body=body,
+                    size=k,
+                    _source=["id"],
+                    preference="_only_local" if self.case_config.number_of_shards == 1 else None,
+                    routing=self.routing_key,
+                )
+                try:
+                    return [int(h["_source"]["id"]) for h in resp["hits"]["hits"]]
```
XuanYang-cn, line 470: For serverless we now read benchmark IDs from `_source.id`, but `prepare_filter()` still builds numeric filters on `self.id_col_name` (`_id`) below. That means filtered AOSS cases such as `NewIntFilterPerformanceCase` or `CloudPayloadSearchCase --filter-rate` will not filter on the IDs we inserted. Please switch the serverless `NumGE` filter to the stored `id` field, and make sure the mapping stores it as a numeric field so range queries work.

vectordb_bench/backend/clients/aws_opensearch/config.py
```diff
@@ -16,8 +16,60 @@ class AWSOpenSearchConfig(DBConfig, BaseModel):
...
+        if self.is_serverless:
+            return self._serverless_config()
+        return self._standard_config()
+
+    def _serverless_config(self) -> dict:
+        """Configuration for OpenSearch Serverless using AWS SigV4 authentication."""
+        log.info(f"Configuring OpenSearch Serverless - Host: {self.host}, Region: {self.aws_region}")
+
+        try:
+            import boto3
```
XuanYang-cn, line 32: This new serverless path imports `boto3` and `requests_aws4auth`, but the package extra for OpenSearch still only installs `opensearch-py`, and the README only calls out `requests-aws4auth`. A clean `vectordb-bench[opensearch]` install will fail here before it can connect. Please add the required serverless dependencies to the relevant extra (or a documented serverless extra) and include `boto3` in the docs.

norrishuang:
Thanks for the review, @XuanYang-cn! I've addressed all three comments in 57f7448:

1. **Multi-client insert path**: Serverless now always routes through the single-client insert path (`insert_embeddings()` short-circuits when serverless), so AOSS loads no longer hit the multi-client path that sends a custom `_id`. This keeps the no-custom-`_id` / `_source.id` / small-batch behavior for all serverless loads regardless of `number_of_indexing_clients`.

2. **Filters**: `prepare_filter()` now builds the `NumGE` range on the stored `id` field for serverless (instead of `_id`), and the mapping stores `id` as a numeric `long` field so range queries work. Filtered cases like `NewIntFilterPerformanceCase` will now filter on the inserted IDs correctly.

3. **Dependencies**: Added `boto3` and `requests-aws4auth` to the `opensearch` extra in `pyproject.toml` (and to `install/requirements_py3.11.txt`), and updated the README serverless prerequisites to reference `pip install 'vectordb-bench[opensearch]'` and mention `boto3`. A clean `vectordb-bench[opensearch]` install now includes everything needed for serverless.

PTAL when you get a chance.

/assign @xuanyang-cn

XuanYang-cn (approved):
/lgtm

PR #803, opened 2026-06-22
feat(runner): add configurable --serial-cooldown between concurrent a…

zpatronus:
…nd serial search phases

The serial search phase measures single-query latency. When it launches immediately after the concurrent phase, the end-to-end path may still be in a saturated state, affecting the accuracy of serial latency especially p99/p95. Add a --serial-cooldown parameter (default 0, in seconds, supports decimals) so backends that need it can opt in.


zpatronus:
This is a followup of #801 . Instead of hardcoding a 10s cooldown, this makes the cooldown time configurable so backends who need it can opt in.

XuanYang-cn (approved):

PR #804, opened 2026-06-22
feat(volc_mysql): add VolcMySQL backend with HNSW vector index support

FishMage:
Add a vector database backend for VolcMySQL ([Volcano Engine MySQL](https://www.volcengine.com/docs/6313/1978527?lang=en) with native VECTOR type and HNSW vector index), connecting via mysql-connector-python over the MySQL wire protocol.

Components:
- volc_mysql.py: VectorDB implementation with VECTOR(dim) table, bulk LOAD DATA LOCAL INFILE loading, and CREATE VECTOR INDEX via SECONDARY_ENGINE_ATTRIBUTE (algorithm/distance/m/ef_construction and optional quant_algorithm/quant_type/refine_type for SQ/PQ/RaBitQ)
- config.py: DBConfig (host/port/user/password) and VolcMySQLHNSWConfig (m/ef_search/ef_construction + quantization params)
- cli.py: Click command `VolcMySQLHNSW`

Registration:
- Add VolcMySQL to the DB enum in backend/clients/__init__.py with lazy imports for init_cls, config_cls, and case_config_cls
- Register VolcMySQLHNSW CLI command in cli/vectordbbench.py
- Add volc_mysql optional dependency in pyproject.toml

Binary VECTOR path with auto-probe fallback:
- Vectors are sent as raw little-endian float32 bytes (UNHEX on load, the `_binary` introducer on query), avoiding to_vector() text parsing and Python str() formatting; `_binary` literals stay constant-foldable so the HNSW index scan is preserved
- init() probes once per connection (session-local TEMPORARY TABLE) whether the server accepts the binary path and transparently falls back to the to_vector() text path -- for both insert and query -- when it does not. VDB_BINARY_VEC overrides the probe (1=force binary, 0=force text)

Tests: tests/test_volc_mysql_encoder.py covers binary and text TSV encoding.


XuanYang-cn:
@FishMage Please fix the ut to pass the github actions

XuanYang-cn:
@FishMage THe code check still fails, PTAL

XuanYang-cn (approved):
/lgtm

PR #813, opened 2026-07-14
feat: add insert batch and streaming rate controls

jamesgao-jpg:
## Summary

This upstream PR supersedes [jamesgao-jpg/VectorDBBench#3](https://github.com/jamesgao-jpg/VectorDBBench/pull/3).

- add task-level `insert_batch_size` across CLI, REST, frontend, datasets, runners, metrics, and load-reuse identity
- keep `streaming_insert_rate` case-specific and validate that it is positive, at least the batch size, and divisible by it
- replace `NUM_PER_BATCH` and the Cloud Insert-specific batch option with the common task setting
- separate MemoryDB pipeline buffering as `--memorydb-pipeline-batch-size`, while accepting the legacy serialized field
- preserve semantic qrel documents in capped FTS corpora by preparing the selected corpus before timed insertion; unfiltered loads still omit `filter_id`

## Compatibility

- new CLI options: `--insert-batch-size` and `--streaming-insert-rate`
- removed CLI option: `--cloud-insert-batch-size`
- old result files migrate Cloud Insert `custom_case.batch_size` and metric `num_per_batch` values into `TaskConfig.insert_batch_size`
- backend-native request caps and sub-batching remain unchanged
- capped FTS datasets use a temporary prepared corpus so qrel selection does not require a full raw-corpus scan during timed insertion

## Verification

- `make lint`
- `make unittest`
- 86 focused contract, CLI, runner, frontend, rate, reuse, and Cloud Insert tests
- `python -m pytest -q tests/test_fts_dataset.py` (`32 passed`)
- `python -m pytest -q tests/test_frontend_run_settings.py` (`5 passed`)
- Milvus 2.6.20 functional probe: configured batch size 7 produced three 7-row insert calls, inserted 21/21 rows, completed indexing, and returned search results
- server-hosted Streamlit smoke test: both insert controls rendered and browser console had no errors

Signed-off-by: jamesgao-jpg <james.gao@zilliz.com>


jamesgao-jpg (commented):
Review withdrawn. Feedback is retained locally.

vectordb_bench/backend/dataset.py
```diff
@@ -1046,7 +1051,8 @@ def prepare(
                     raise ValueError(msg)  # noqa: TRY301
 
                 self.required_doc_ids = {doc_id for qrels in self.gt_data for doc_id in qrels}
-                self.selected_doc_ids = self._build_selected_doc_ids()
+                is_filtered = filters is not None and filters.type != FilterOp.NonFilter
+                self.selected_doc_ids = self._build_selected_doc_ids() if is_filtered else None
```
XuanYang-cn: For unfiltered capped FTS runs, setting `selected_doc_ids` to `None` makes the iterator load only the first `data.size` corpus documents while `gt_data` still contains every positive qrel. Any qrel document after the cap is therefore impossible to retrieve, which systematically lowers recall/NDCG/MRR. Please keep the qrel-preserving document selection for unfiltered runs too; omitting `filter_id` does not require bypassing `_build_selected_doc_ids()`.

vectordb_bench/frontend/config/dbCaseConfigs.py
```diff
@@ -240,9 +240,10 @@ def generate_custom_streaming_case() -> CaseConfig:
     ),
     ConfigInput(
         label=CaseConfigParamType.insert_rate,
+        displayLabel="Streaming Insert Rate",
         inputType=InputType.Number,
-        inputConfig=dict(step=100, min=100, max=4_000, value=200),
-        inputHelp="fixed insertion rate (rows/s), must be divisible by 100",
+        inputConfig=dict(step=100, min=100, max=MAX_STREAMLIT_INT, value=500),
```
XuanYang-cn: This hard-coded minimum rejects valid configurations. For example, `insert_batch_size=10` and `streaming_insert_rate=50` satisfy all documented constraints and are accepted by CLI/REST, but the frontend cannot represent them. Please use a minimum of 1 and let the existing batch/rate validation enforce the relationship.

XuanYang-cn (changes requested):
Please address the two inline findings before merging.

vectordb_bench/backend/dataset.py
jamesgao-jpg: Fixed in [64771cd](https://github.com/zilliztech/VectorDBBench/commit/64771cda9d90e6be4f9f57f4cfa48cb3281ee5ed). Capped FTS datasets now materialize a qrel-preserving corpus during untimed dataset preparation, and both filtered and unfiltered insertion stream that prepared corpus. This keeps every positive-qrel document retrievable while unfiltered loads still omit `filter_id`, without requiring a full raw-corpus selection scan during timed insertion. Added regression coverage for a qrel document beyond the cap and for filtered/unfiltered prepared-corpus iteration; `tests/test_fts_dataset.py` passes all 32 tests, with focused Ruff and Black checks also passing.

vectordb_bench/frontend/config/dbCaseConfigs.py
jamesgao-jpg: Fixed in [a22c48a](https://github.com/zilliztech/VectorDBBench/commit/a22c48a66e1a91b6a37e4a1a8fd064b3ab86bcd8). The frontend streaming insert-rate minimum is now 1, while the existing validation continues to enforce the relationship between `streaming_insert_rate` and `insert_batch_size`. I also added a regression test covering `streaming_insert_rate=50` with `insert_batch_size=10`; `python -m pytest -q tests/test_frontend_run_settings.py` passes (5 tests).

XuanYang-cn (approved):

PR #816, opened 2026-07-14
Add hyperspacedb client

YARlabs:
### Summary

We'd like to contribute a new backend client for **HyperspaceDB**, a vector database written in Rust (project: YAR.INK), and — if the maintainers are open to it — have it evaluated on the standard leaderboard alongside the existing clients (Milvus, Qdrant, Weaviate, PgVector, etc.).

To keep the comparison fair and easy to review, **this proposal is scoped to standard, well-understood ground**: HNSW-based ANN search under the same **L2 / Cosine** metrics are already used for every other client in the leaderboard.

### What we are explicitly *not* proposing (yet)

HyperspaceDB also implements non-Euclidean capabilities — a Lorentz/Poincaré hyperbolic index and an experimental wave-diffusion search mode. We are intentionally leaving all of that out of this PR. Reasons:

- Those modes don't map onto existing `MetricType`/case definitions in  VectorDBBench, and we don't want to ask maintainers to build new infrastructure just to evaluate us.
- We want the first data point to be a clean, apples-to-apples comparison on the exact terms the benchmark already tests everyone on.
- Hyperbolic/wave benchmarking can be a separate follow-up proposal once (if) the maintainers are interested, with their own metric/case definitions discussed on their own merits.

So, this PR tests HyperspaceDB purely as an HNSW vector index over L2/Cosine, competing on the same field as everyone else.

### What's included in the PR

- `vectordb_bench/backend/clients/hyperspacedb/`
  - `config.py` — `HyperspaceDBConfig(DBConfig)`, `HyperspaceDBIndexConfig(DBCaseConfig)` (HNSW params: `m`, `ef_construction`, `ef_search`; metric restricted to `L2`/`COSINE` for this PR)
  - `hyperspacedb.py` — `HyperspaceDB(VectorDB)` implementing the standard insert/search/optimize interface
- Registration of `HyperspaceDB` in the `DB` enum and `db_config`/`db_case_config`
- CLI command in `vectordb_bench/cli/vectordbbench.py`
- `pyproject.toml`: new `hyperspacedb` extra
- Tests under `tests/test_hyperspacedb.py` following the existing client test pattern
- Docs: a short usage section (connection params, install instructions, example `vectordbbench hyperspacedb --help` command), matching the style used for other clients

### What we're asking from maintainers

1. **Code review** of the client implementation against your `VectorDB` interface conventions.
2. If the client is accepted, we understand the results on the public leaderboard require running on your standard reference client hardware (8 vCPU / 32GB, same region as the server) for consistency with the other entries. We're happy to provide a hosted or Docker-deployable instance of HyperspaceDB for that run, or to run it ourselves against your published methodology and submit results for your verification - whichever is easier on your side.
3. We are not asking for special-casing or a separate "non-Euclidean" leaderboard track in this PR — just a seat at the existing L2/Cosine table.

### Why we think this is worth your time

We believe HyperspaceDB's HNSW implementation performs competitively against the databases currently on the leaderboard, even under the standard metrics they were designed around — that's the specific claim this PR is meant to let your benchmark verify (or refute) independently, rather than something we're asserting without your test.

### Checklist

- [ ] Client implements full `VectorDB` interface (insert, search, optimize, `ready_to_load`, `need_normalize_cosine`)
- [ ] `db_config` / `db_case_config` reviewed against project conventions
- [ ] CLI command added and documented
- [ ] `pyproject.toml` extra added (`hyperspacedb`)
- [ ] Tests included and passing
- [ ] Docs updated
- [ ] Open to running on maintainers' reference hardware for leaderboard inclusion

### DIFF
--- a/README.md
+++ b/README.md
@@ -68,6 +68,7 @@ All the database client supported
 | lindorm                  | `pip install vectordb-bench[lindorm]`       |
 | volc_mysql               | `pip install vectordb-bench[volc_mysql]`    |
 | adbpg                    | `pip install vectordb-bench[adbpg]`         |
+| hyperspacedb             | `pip install vectordb-bench[hyperspacedb]`  |
 
 ### Run
 
--- a/vectordb_bench/backend/clients/hyperspacedb/__init__.py
+++ b/vectordb_bench/backend/clients/hyperspacedb/__init__.py
@@ -1,0 +1,4 @@
+from .config import HyperspaceDBConfig, HyperspaceDBIndexConfig
+from .hyperspacedb import HyperspaceDB
+
+__all__ = ["HyperspaceDBConfig", "HyperspaceDBIndexConfig", "HyperspaceDB"]
--- a/vectordb_bench/backend/clients/hyperspacedb/cli.py
+++ b/vectordb_bench/backend/clients/hyperspacedb/cli.py
@@ -1,0 +1,57 @@
+from typing import Annotated, Unpack
+
+import click
+from pydantic import SecretStr
+
+from vectordb_bench.backend.clients import DB
+from vectordb_bench.cli.cli import (
+    CommonTypedDict,
+    cli,
+    click_parameter_decorators_from_typed_dict,
+    run,
+)
+
+DBTYPE = DB.HyperspaceDB
+
+
+class HyperspaceDBTypeDict(CommonTypedDict):
+    host: Annotated[
+        str,
+        click.option("--host", type=str, help="HyperspaceDB host", default="localhost:50051"),
+    ]
+    api_key: Annotated[
+        str | None,
+        click.option("--api-key", type=str, help="HyperspaceDB API Key", required=False),
+    ]
+    m: Annotated[
+        int,
+        click.option("--m", type=int, help="HNSW Maximum Neighbors", default=16),
+    ]
+    ef_construction: Annotated[
+        int,
+        click.option("--ef-construction", type=int, help="HNSW efConstruction", default=100),
+    ]
+    ef_search: Annotated[
+        int,
+        click.option("--ef-search", type=int, help="HNSW efSearch", default=100),
+    ]
+
+
+@cli.command()
+@click_parameter_decorators_from_typed_dict(HyperspaceDBTypeDict)
+def HyperspaceDB(**parameters: Unpack[HyperspaceDBTypeDict]):
+    from .config import HyperspaceDBConfig, HyperspaceDBIndexConfig
+
+    run(
+        db=DBTYPE,
+        db_config=HyperspaceDBConfig(
+            host=parameters["host"],
+            api_key=SecretStr(parameters["api_key"]) if parameters["api_key"] else None,
+        ),
+        db_case_config=HyperspaceDBIndexConfig(
+            m=parameters["m"],
+            ef_construction=parameters["ef_construction"],
+            ef_search=parameters["ef_search"],
+        ),
+        **parameters,
+    )
--- a/vectordb_bench/backend/clients/hyperspacedb/hyperspacedb.py
+++ b/vectordb_bench/backend/clients/hyperspacedb/hyperspacedb.py
@@ -133,7 +133,20 @@ class HyperspaceDB(VectorDB):
     ) -> list[int]:
         assert self.client is not None, "Please call self.init() before"
         
-        # Build filter if present
+        # Direct gRPC fast-path to bypass Python SDK overhead when no filters are present
+        if not filters:
+            try:
+                from hyperspace.proto import hyperspace_pb2
+                req = hyperspace_pb2.SearchRequest(
+                    vector=query,
+                    top_k=k,
+                    collection=self.collection_name
+                )
+                resp = self.client.stub.Search(req, metadata=self.client.metadata)
+                return [r.id for r in resp.results]
+            except Exception as e:
+                log.warning(f"Fast-path search failed: {e}. Falling back to SDK search.")
+        
         flt = None
         if filters:
             # Map simple filter if supported
--- a/vectordb_bench/backend/clients/hyperspacedb/hyperspacedb.py
+++ b/vectordb_bench/backend/clients/hyperspacedb/hyperspacedb.py
@@ -71,11 +71,31 @@ class HyperspaceDB(VectorDB):
         # Apply search param config
         try:
             self.client.configure(
+                collection=self.collection_name,
                 ef_search=self.case_config.ef_search,
-                collection=self.collection_name
+                ef_construction=self.case_config.ef_construction,
             )
         except Exception as e:
             log.warning(f"Optimize configure error: {e}")
+            
+        # Wait for background indexing to complete using gRPC stats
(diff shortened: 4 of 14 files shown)


XuanYang-cn:
@YARlabs Please fix the github actions

vectordb_bench/backend/clients/hyperspacedb/__init__.py
```diff
@@ -0,0 +1,4 @@
+from .config import HyperspaceDBConfig, HyperspaceDBIndexConfig
+from .hyperspacedb import HyperspaceDB
```
XuanYang-cn, line 2: Please remove this `__init__.py`; the other client directories do not add one, and the existing lazy `DB.init_cls` registration already handles loading the adapter when needed. `vectordbbench.py` imports `hyperspacedb.cli` unconditionally, so this initializer eagerly imports the optional SDK and makes a base install fail with `ModuleNotFoundError: No module named 'hyperspace'` before any command can run.

vectordb_bench/backend/clients/hyperspacedb/hyperspacedb.py
```diff
@@ -0,0 +1,190 @@
...
+            try:
+                client.delete_collection(self.collection_name)
+                time.sleep(0.5)
+            except Exception as e:
+                log.info(f"Failed to drop old collection: {e}")
+                
+        # Create collection
+        try:
+            metric = self.case_config.parse_metric()
+            client.create_collection(
```
XuanYang-cn, line 51: The published SDK's `create_collection()` and `configure()` methods return `False` on RPC errors instead of raising, so this `try` block does not detect most setup failures. Please check both return values and raise when either operation fails; otherwise the benchmark proceeds without a correctly created/configured collection.

vectordb_bench/backend/clients/hyperspacedb/hyperspacedb.py
```diff
@@ -0,0 +1,190 @@
...
+        
+        # Prepare metadata
+        labels_data = kwargs.get("labels_data")
+        if labels_data is not None:
+            metas = [{"id": str(idx), "labels": str(label)} for idx, label in zip(metadata, labels_data)]
+        else:
+            metas = [{"id": str(idx)} for idx in metadata]
+        
+        try:
+            self.client.batch_insert(
```
XuanYang-cn, line 141: `batch_insert()` returns a boolean and returns `False` for RPC/server failures. Its result is ignored here, so a failed batch is reported as `(len(embeddings), None)` and VectorDBBench skips its retry path. Please return an error when the SDK reports `False`.

vectordb_bench/backend/clients/hyperspacedb/hyperspacedb.py
```diff
@@ -0,0 +1,190 @@
...
+        timeout = 3600  # 1 hour max
+        while True:
+            if time.time() - start_time > timeout:
+                log.warning("Indexing timeout reached during optimize")
+                break
+            try:
+                stats = self.client.get_collection_stats(self.collection_name)
+                queue = stats.get("indexing_queue", 0)
+                count = stats.get("count", 0)
+                if queue == 0 and count > 0:
```
XuanYang-cn, line 117: `TaskRunner` passes the expected dataset size through `data_size`, but this considers any positive count ready. That can accept a partial or stale collection as fully loaded. Please require `count >= data_size`; also treat empty stats as an error, because the SDK returns `{}` on RPC failure and the current code then polls until the one-hour timeout.

vectordb_bench/frontend/config/dbCaseConfigs.py
```diff
@@ -3247,6 +3247,19 @@ class FilterType(Enum):
     CaseConfigParamInput_NovaAdaptiveGamma_Adbpg,
 ]
 
+HyperspaceDBLoadConfig = [
+    CaseConfigParamInput_IndexType,
```
XuanYang-cn, line 3251: This generic selector exposes IVF, PQ, GPU, Flat, AUTOINDEX, and other choices, while the adapter always configures HNSW and ignores the selected `index`. A user can therefore publish a result labeled as a different index while actually running HNSW. Please use a HyperspaceDB-specific selector limited to HNSW, or omit the selector.

vectordb_bench/results/HyperspaceDB/result_20260712_2026071214_hyperspacedb.json
```diff
@@ -0,0 +1,794 @@
+{
```
XuanYang-cn, line 1: Please remove this file and the other two HyperspaceDB result files. We do not accept benchmark results generated by contributors into the repository. You are welcome to paste these results in the PR comments for others' reference, but we only merge results tested by maintainers in our own test environment.

vectordb_bench/backend/clients/hyperspacedb/cli.py
```diff
@@ -0,0 +1,57 @@
...
+        str | None,
+        click.option("--api-key", type=str, help="HyperspaceDB API Key", required=False),
+    ]
+    m: Annotated[
+        int,
+        click.option("--m", type=int, help="HNSW Maximum Neighbors", default=16),
+    ]
+    ef_construction: Annotated[
+        int,
+        click.option("--ef-construction", type=int, help="HNSW efConstruction", default=100),
```
XuanYang-cn, line 32: This defaults `ef_construction` to `100`, while `HyperspaceDBIndexConfig` and the UI default it to `200`. The same benchmark therefore changes depending on whether it is started through CLI or UI. Please define and reuse one default.

README.md
```diff
@@ -68,6 +68,7 @@ All the database client supported
 | lindorm                  | `pip install vectordb-bench[lindorm]`       |
 | volc_mysql               | `pip install vectordb-bench[volc_mysql]`    |
 | adbpg                    | `pip install vectordb-bench[adbpg]`         |
+| hyperspacedb             | `pip install vectordb-bench[hyperspacedb]`  |
```
XuanYang-cn, line 71: The PR description promises a short usage section with connection parameters and an example `vectordbbench hyperspacedb --help` command, but this only adds the installation row. Please document `--host`, optional `--api-key`, and a minimal CLI example.

vectordb_bench/backend/clients/hyperspacedb/hyperspacedb.py
```diff
@@ -0,0 +1,190 @@
...
+    @contextmanager
+    def init(self):
+        host = self.db_config.get("host", "localhost:50051")
+        api_key = self.db_config.get("api_key")
+        self.client = HyperspaceClient(host, api_key=api_key, pool_size=64)
+        yield
+        self.client.close()
+        self.client = None
+        
+    def ready_to_search(self) -> bool:
```
XuanYang-cn, line 77: This method is not part of the current `VectorDB` readiness API and always returns `None`. Meanwhile, the PR checklist promises a readiness implementation. Please remove this dead method or implement the current readiness hook, such as `poll_insert_readiness()`, and update the checklist accordingly.

XuanYang-cn (changes requested):
Please address the inline review comments.

PR #818, opened 2026-07-16
feat(cli): add common note support

jamesgao-jpg:
## What this changes

- Adds shared `--note` and `--note-file` options to backend CLI commands.
- Resolves the supplied context centrally into `task_config.db_config.note`, which is already serialized into each result JSON.
- Ensures Endee's custom CLI runner also resolves note files before constructing its task.
- Rejects conflicting options, empty note files, and non-UTF-8 note files.
- Preserves Pinecone's existing `--note` behavior through the common implementation.
- Documents how to attach structured or multiline benchmark context.

## Compatibility

- Existing CLI commands remain unchanged when neither option is supplied.
- This does not change the result schema because `DBConfig.note` already exists.
- Backend connection behavior is unchanged: provider clients receive their existing explicit `DBConfig.to_dict()` connection fields, not the note metadata.

## Verification

- `make lint`
- `make unittest` — 1 passed
- `python -m pytest tests/test_cli_note.py -q` — 7 passed
- `python -m pytest tests/test_milvus_zilliz_cli.py -q` — 2 passed
- Verified `--note` and `--note-file` help output for Zilliz Cloud and Pinecone commands.

### DIFF
--- a/vectordb_bench/cli/cli.py
+++ b/vectordb_bench/cli/cli.py
@@ -66,6 +66,20 @@ def click_get_defaults_from_file(ctx, param, value):  # noqa: ANN001, ARG001
     return value
 
 
+def resolve_db_note(note: str, note_file: Path | None) -> str:
+    if note and note_file is not None:
+        raise click.UsageError("--note and --note-file cannot be used together")
+    if note_file is None:
+        return note
+    try:
+        content = note_file.read_text(encoding="utf-8").rstrip("\r\n")
+    except UnicodeDecodeError as e:
+        raise click.BadParameter("Note file is not valid UTF-8", param_hint="--note-file") from e
+    if not content.strip():
+        raise click.BadParameter("Note file is empty", param_hint="--note-file")
+    return content
+
+
 def click_parameter_decorators_from_typed_dict(
     typed_dict: type,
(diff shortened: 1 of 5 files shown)


vectordb_bench/cli/cli.py
```diff
@@ -66,6 +66,20 @@ def click_get_defaults_from_file(ctx, param, value):  # noqa: ANN001, ARG001
     return value
 
 
+def resolve_db_note(note: str, note_file: Path | None) -> str:
+    if note and note_file is not None:
```
XuanYang-cn: `resolve_db_note()` checks the value truthiness, so an explicitly supplied empty value bypasses the mutual-exclusion check: `--note "" --note-file README.md --dry-run` succeeds and uses the file. This contradicts the documented mutually-exclusive behavior. Please enforce exclusivity based on whether each Click option was supplied, and add this empty-inline case to the regression test.

jamesgao-jpg: Fixed in 9469c8e. resolve_db_note() now checks Click parameter sources, so an explicitly supplied empty --note conflicts with --note-file. Added regression coverage for --note "" --note-file ..., and all four PR CI jobs pass.

PR #823, opened 2026-07-21
feat(clients): add Telys connector (on-device, Algenta-powered vector engine)

angelatgithub:
## What
Adds a VectorDBBench client for **Telys** — an on-device, Algenta-powered vector-execution layer that accelerates *filtered* search by physical layout: the partition key is stored as a contiguous slice, so `label == X` reads one block instead of scanning + masking the whole table. The connector talks to a running `telys serve` (TCP) via the public [`telys`](https://pypi.org/project/telys/) SDK, so it's a fair server-vs-server measurement. It partitions by the label field, so the **label-filter case (`StrEqual`) maps to Telys's single-key contiguous-partition wedge**.

## Changes
- `backend/clients/telys/{telys.py,config.py,cli.py,__init__.py}` — `VectorDB` connector + config + CLI
- Registered in `backend/clients/__init__.py` (DB enum + `init_cls`/`config_cls`/`case_config_cls`) and `cli/vectordbbench.py`
- `pyproject.toml`: `telys` extra (`pip install vectordb-bench[telys]`, `telys>=0.1.0b3`); README supported-clients table updated

## Supported filters
- `NonFilter` and `StrEqual` (label==value — the wedge)
- `NumGE` (int-range) is intentionally **not** declared — Telys `Eq` is equality, not a range
- Cosine/IP datasets (`need_normalize_cosine=True`)

## Validation
Run end-to-end through the real `CaseRunner` against a live `telys serve`:
- **Cohere-100K, 768-d cosine, label-filter @ 1%**: recall **1.0000**, ndcg **1.0000**, serial p99 **1.5 ms**, **4099 qps @ concurrency 8**
- Custom local fixture (unfiltered + label filters @ 50% and 2%): recall **1.000**; concurrent multiprocessing search (each worker its own connection); no crashes

## Run
```
pip install vectordb-bench[telys]
telys serve --host 0.0.0.0 --port 9099 --access-token "$TELYS_ACCESS_TOKEN"
vectordbbench telys --host 127.0.0.1 --port 9099 --access-token "$TELYS_ACCESS_TOKEN" ...
```

## Notes for maintainers
- Inserts go over the SDK's length-prefixed JSON wire (search latency — the headline metric — is unaffected); load throughput is thus conservative vs binary-protocol clients.
- Happy to add a curated-board entry — could you point me at the process for inclusion in the `zilliz.com/vdbbench-leaderboard` set?


angelatgithub:
Closing to resubmit cleanly with correct commit authorship (angelatgithub). Reopening shortly.

PR #832, opened 2026-08-03
feat(adbpg): add ADBPG generic search configuration

swhhy:
## Summary

- Add generic ADBPG configuration injection for CREATE INDEX reloptions and INCLUDE columns, search-session GUCs, post-index setup SQL, and post-build index reloptions.
- Add NOVA autotune execution after index build and setup SQL, with polling until the asynchronous task exits and all expected configuration rows are present.
- Require autotune `topk` to be configured explicitly, independently of benchmark `k`; both `topk` and `target_recall` support scalar and `ARRAY[...]` values.
- Preserve setup SQL verbatim and keep commas inside configuration values.
- Add a Cohere-1M NOVAMR example covering import, index build, autotune, and top-10 search. The example links to the NOVA performance guide.
- Document the new ADBPG options and run the ADBPG configuration regressions in pull-request CI.

## Example configuration

[`vectordb_bench/config-files/adbpg_cohere1m_autotune.yml`](https://github.com/zilliztech/VectorDBBench/blob/main/vectordb_bench/config-files/adbpg_cohere1m_autotune.yml)

## Validation

- `make lint`
- `pytest tests/test_adbpg_search_config.py -q` — 12 passed


vectordb_bench/backend/clients/adbpg/adbpg.py
```diff
@@ -191,6 +196,70 @@ def _drop_index(self):
         self.cursor.execute(drop_index_sql)
         self.conn.commit()
 
+    @staticmethod
+    def _quote_identifier(name: str) -> str:
+        return '"' + name.replace('"', '""') + '"'
+
+    def _render_setup_sql(self, statement: str) -> str:
+        """Resolve stable benchmark placeholders in explicitly injected SQL."""
+        return Template(statement).safe_substitute(
```
yhmo: Medium ---- `string.Template` treats `$$` as an escape and collapses it to a single `$`. That breaks a common PostgreSQL setup statement such as `DO $$ BEGIN ... END $$;` before it reaches the server, even though `--setup-sql` is advertised as accepting SQL. The regression test only covers a named dollar tag (`$body$`), which `safe_substitute` happens to leave alone. Please preserve ordinary PostgreSQL dollar-quoted bodies while replacing only the three documented placeholders, and add a `DO $$...$$` case.

vectordb_bench/backend/clients/adbpg/config.py
```diff
@@ -46,6 +46,7 @@ def to_dict(self) -> dict:
 
 class AdbpgIndexConfig(BaseModel, DBCaseConfig):
     metric_type: MetricType | None = None
+    benchmark_topk: int = 100
```
yhmo: Medium ---- This duplicates the benchmark k with an independent default, but only the direct CLI path copies `parameters["k"]` into it. Batch/UI/REST tasks keep the real value in `TaskConfig.case_config.k`, so a run configured with (for example) k=10 still expands `$topk` to 100 unless the user knows to set this internal field separately. That silently tunes the index/setup SQL for a different workload. Please propagate `CaseConfig.k` to the ADBPG client for every entry point (or pass it as a normal client argument) instead of maintaining a second default.

vectordb_bench/backend/clients/adbpg/options.py
```diff
@@ -0,0 +1,36 @@
+from typing import Any
+
+import click
+
+
+def _split_config_items(values: tuple[str, ...]) -> list[str]:
+    """Expand repeatable and comma-separated CLI/YAML values."""
+    return [part.strip() for value in values for part in str(value).split(",") if part.strip()]
```
yhmo: vectordb_bench/backend/clients/adbpg/options.py line:8
Medium ---- `_split_config_items` always splits each option value on commas before parsing `name=value`, so valid comma-valued GUCs such as `search_path=foo,bar` cannot be expressed: `bar` is treated as another item and rejected. Because this option is advertised as generic session-GUC support, please either remove the comma shorthand or add an escaping/quoting rule that preserves commas inside values, with a regression test.

vectordb_bench/backend/clients/adbpg/adbpg.py
swhhy: Thanks. I removed placeholder substitution entirely. setup_sql is now executed verbatim, and the regression test covers both DO $$ ... $$ and literal $topk, $table, and $index strings.

vectordb_bench/backend/clients/adbpg/config.py
swhhy: Thanks. ADBPG now receives TaskConfig.case_config.k from CaseRunner, so CLI, batch, UI, and REST tasks all use the actual workload topK. Autotune derives topk from that value and no longer maintains a second default.

vectordb_bench/backend/clients/adbpg/options.py
swhhy: Thanks. Repeatable configuration options are no longer split on commas. Each option occurrence represents one item and is split only at the first =, so values such as search_path=foo,bar are preserved. A regression test was added for this case.

vectordb_bench/backend/clients/adbpg/config.py
swhhy: Correction to my previous reply: the final implementation no longer derives autotune topk from CaseConfig.k. autotune_param now requires an explicit topk, independent of benchmark k. Both topk and target_recall accept either a scalar or ARRAY[...] value, so one autotune run can cover multiple target combinations.

swhhy:
/assign @XuanYang-cn   @yhmo 
Hi, I’ve addressed all three review comments and successfully run make lint. Could you please approve the workflow run for this PR🙏? Thanks!!!

PR #836, opened 2026-08-05
fix query_mode supoort milvus db type

frankleaf:
I noticed that your workflow specifies a ZILLIZ_CLOUD_DEFAULT_TOPK_LIMIT restriction and then enters large_topk search mode. I think this is also necessary for open-source Milvus.


XuanYang-cn:
/assign @jamesgao-jpg 
PTAL

jamesgao-jpg:
Hi, thanks for the contribution. Looks mostly good to me! When you have time please take a look at the following trivias and patch them up and I will merge it. 

1. Update tests/test_large_topk_case.py. The existing test explicitly expects DB.Milvus with K=100,000 to remain unchanged, so it now conflicts with this change. Please add coverage confirming Milvus selects query_mode=large_topk, while K <= 16,384 remains in default mode.

2. Update the README and large-TopK release notes. They currently state that only Zilliz Cloud is configured automatically and that Milvus remains unchanged.

3. Document that this behavior requires Milvus 2.6.14 or later, where the query_mode=large_topk collection property became available.

frankleaf:
@jamesgao-jpg Thanks! I've added the tests and README as requested. Please take a look

PR #842, opened 2026-08-10
feat(turso): add Turso backend

uwu-420:
Adding a backend for https://github.com/tursodatabase/turso


vectordb_bench/backend/clients/turso/turso.py
```diff
@@ -0,0 +1,172 @@
...
+        expected_shape = (len(metadata), self.dim)
+        if vectors.shape != expected_shape:
+            msg = f"Expected embeddings with shape {expected_shape}, got {vectors.shape}"
+            return 0, ValueError(msg)
+        with self._operation_lock:
+            connection = self._connection()
+            try:
+                connection.executemany(
+                    "INSERT INTO vectors(id, embedding) VALUES (?, ?)",
+                    [(int(row_id), vector.tobytes()) for row_id, vector in zip(metadata, vectors, strict=True)],
```
yhmo, line 111: vectordb_bench/backend/clients/turso/turso.py line:111
High ---- Turso's vector_distance_* functions require vector32/vector64-encoded blobs, but this inserts raw NumPy float bytes directly. Those blobs are not valid Turso vectors, so every subsequent distance query will fail validation (or misinterpret the data) once the backend is used. Please encode each vector with Turso's vector32 representation before storing it.

uwu-420: Thanks for checking this. For the pinned `pyturso` 0.7.2 release, Float32 is the special case whose binary representation is exactly the raw f32 payload. [`Vector::vector_type`](https://github.com/tursodatabase/turso/blob/v0.7.2/core/vector/vector_types.rs#L33-L37) treats every even-length blob as `Float32Dense`, and [`vector_serialize`](https://github.com/tursodatabase/turso/blob/v0.7.2/core/vector/operations/serialize.rs#L6-L9) emits `Float32Dense` without a type byte.

I verified this against the pinned wheel: `vector32('[1,2,3]')` returns the same 12 bytes as `np.asarray([1,2,3], dtype='<f4').tobytes()`, and direct bound blobs match `vector32(...)` for cosine, L2, and dot distance. The adapter tests also exercise insertion and search through this path. Calling `vector32(?)` would parse and reserialize the same valid bytes, so no additional encoding appears to be missing here. Please let me know if there is another compatibility constraint I have overlooked.

yhmo: Thanks for the clarification. I agree that no change is needed here.

PR #843, opened 2026-08-11
Make AOSS insert batch size configurable

norrishuang:
## What changed

- Use the existing `NUM_PER_BATCH` configuration for OpenSearch Serverless bulk requests instead of a hard-coded batch size of 100.
- Keep 100 as the default while allowing callers to override it through the environment.
- Document the environment variable and add a regression test for AOSS bulk chunking.

## Testing

```shell
PYTHONPATH=`pwd` python3 -m pytest -q tests/test_aws_opensearch.py
PYTHONPATH=`pwd` python3 -m black vectordb_bench
PYTHONPATH=`pwd` python3 -m ruff check vectordb_bench --fix
PYTHONPATH=`pwd` python3 -m black vectordb_bench --check
PYTHONPATH=`pwd` python3 -m ruff check vectordb_bench
```

### DIFF
--- a/vectordb_bench/backend/clients/aws_opensearch/aws_opensearch.py
+++ b/vectordb_bench/backend/clients/aws_opensearch/aws_opensearch.py
@@ -5,6 +5,7 @@ from contextlib import contextmanager
 
 from opensearchpy import OpenSearch
 
+from vectordb_bench import config
 from vectordb_bench.backend.filter import Filter, FilterOp
 
 from ..api import VectorDB
@@ -237,8 +238,8 @@ class AWSOpenSearch(VectorDB):
         num_clients = self.case_config.number_of_indexing_clients or 1
         log.info(f"Number of indexing clients from case_config: {num_clients}")
 
-        # OpenSearch Serverless requires the single-client path: it does not support
-        # custom _id and needs the benchmark id stored in _source with small batches.
+        # OpenSearch Serverless requires the single-client path because it does not
+        # support custom _id and needs the benchmark id stored in _source.
         if self._is_serverless:
             log.info("Using single client for data insertion (OpenSearch Serverless)")
             return self._insert_with_single_client(embeddings, metadata, labels_data)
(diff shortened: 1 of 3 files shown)


vectordb_bench/backend/clients/aws_opensearch/aws_opensearch.py
```diff
@@ -256,7 +257,7 @@ def _insert_with_single_client(
         labels_data: list[str] | None = None,
     ) -> tuple[int, Exception]:
         embeddings_list = list(embeddings)
-        batch_size = 100 if self._is_serverless else len(embeddings_list)
+        batch_size = config.NUM_PER_BATCH if self._is_serverless else len(embeddings_list)
```
yhmo, line 260: vectordb_bench/backend/clients/aws_opensearch/aws_opensearch.py line:260
Medium ---- `NUM_PER_BATCH` comes directly from an environment variable without a positivity check. If it is set to `0`, `range(..., batch_size)` raises `ValueError: range() arg 3 must not be zero` before any request; a negative value makes the range empty and returns `(0, None)`, falsely reporting a successful load. Please validate this setting (or clamp/reject non-positive values) before using it so operator misconfiguration cannot crash or silently skip ingestion.

jamesgao-jpg:
Hi, thank you for the contribution! FYI we have a broader change to make num per batch a CLI variable coming in https://github.com/zilliztech/VectorDBBench/pull/813. Once this pr is merged, 813 will reflect upon this by controlling batch behavior through CLI.

vectordb_bench/backend/clients/aws_opensearch/aws_opensearch.py
norrishuang: Fixed in fc5b23e. OpenSearch Serverless now rejects non-positive NUM_PER_BATCH values with a clear ValueError before constructing the range, and tests cover both zero and negative values.

PR #846, opened 2026-08-12
feat: add ivf_pq support to milvus

FantasqueX:
I don't see similar tests around so not add tests. If you want, I can add them.

### DIFF
--- a/vectordb_bench/backend/clients/milvus/cli.py
+++ b/vectordb_bench/backend/clients/milvus/cli.py
@@ -335,6 +335,52 @@ def MilvusIVFSQ8(**parameters: Unpack[MilvusIVFFlatTypedDict]):
     )
 
 
+class MilvusIVFPQTypedDict(MilvusIVFFlatTypedDict, MilvusRefineTypedDict):
+    nbits: Annotated[
+        int,
+        click.option(
+            "--nbits",
+            type=int,
+            required=True,
+        ),
+    ]
+    pq_m: Annotated[
+        int,
+        click.option(
+            "--pq-m",
+            type=int,
+            default=32,
+            show_default=True,
+            help="Number of PQ sub-quantizers; the vector dim must be divisible by it. "
+            "Higher = finer quantization and better recall, but larger codes.",
+        ),
+    ]
+
+
+@cli.command()
+@click_parameter_decorators_from_typed_dict(MilvusIVFPQTypedDict)
+def MilvusIVFPQ(**parameters: Unpack[MilvusIVFPQTypedDict]):
+    from .config import IVFPQConfig
+
+    run(
+        db=DBTYPE,
+        db_config=_build_milvus_config(parameters),
+        db_case_config=_with_partition_key(
+            IVFPQConfig(
+                nlist=parameters["nlist"],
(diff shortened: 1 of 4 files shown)


vectordb_bench/backend/clients/milvus/cli.py
```diff
@@ -335,6 +335,52 @@ def MilvusIVFSQ8(**parameters: Unpack[MilvusIVFFlatTypedDict]):
...
+@cli.command()
+@click_parameter_decorators_from_typed_dict(MilvusIVFPQTypedDict)
+def MilvusIVFPQ(**parameters: Unpack[MilvusIVFPQTypedDict]):
+    from .config import IVFPQConfig
+
+    run(
+        db=DBTYPE,
+        db_config=_build_milvus_config(parameters),
+        db_case_config=_with_partition_key(
+            IVFPQConfig(
```
yhmo, line 369: vectordb_bench/backend/clients/milvus/cli.py line:369
Low ---- This adds a new user-facing command and manually maps seven IVF_PQ/refine options into the Milvus config, but no test exercises this path. The repository already has `tests/test_milvus_zilliz_cli.py` with a `CliRunner`/captured-`run` seam; please add a case there that invokes `MilvusIVFPQ` and asserts both the index and search payloads, so option-name/default regressions do not silently change the benchmark request.

FantasqueX: I have added a test case.

yhmo: The issue is confirmed fixed.

vectordb_bench/frontend/config/dbCaseConfigs.py
```diff
@@ -794,7 +794,13 @@ class CaseConfigInput(BaseModel):
     inputHelp="Whether refined data is reserved during index building.",
     inputConfig={"options": [True, False]},
     isDisplayed=lambda config: config.get(CaseConfigParamType.IndexType, None)
-    in [IndexType.HNSW_SQ.value, IndexType.HNSW_PQ.value, IndexType.HNSW_PRQ.value, IndexType.IVF_RABITQ.value],
+    in [
+        IndexType.HNSW_SQ.value,
+        IndexType.HNSW_PQ.value,
+        IndexType.HNSW_PRQ.value,
+        IndexType.IVF_RABITQ.value,
+        IndexType.IVFPQ.value,
```
jamesgao-jpg, line 802: [P1] Preserve the existing non-refined IVF_PQ default

`CaseConfigParamInput_Refine` uses `options: [True, False]`, and `inputWidget` passes those options to `st.selectbox` without an explicit index, so `True` is selected by default. Adding `IVFPQ` here therefore enables FP32 refinement for an untouched frontend IVF_PQ run even though `IVFPQConfig` defines `refine=False` and the previous frontend behavior was non-refined. This silently changes the benchmark workload and makes default results incomparable. Please give IVF_PQ a selector that defaults to `False` without changing the existing HNSW/IVF_RABITQ defaults, and add a focused frontend-default regression test.

jamesgao-jpg (commented):
One benchmark-default compatibility issue found.

PR #849, opened 2026-08-14
fix: route insert-runner thread safety through a client capability

### LINKED ISSUE #810: Fixed-rate insert runner can share non-thread-safe DB clients
VectorDBBench has generic runner code that decides database behavior by checking specific DB names. In `RatedMultiThreadingInsertRunner.send_insert_task()`, the runner special-cases only `PgVector`, `Doris`, and `SeekDB` for per-thread client copy/reset/init behavior.

This is fragile because thread-safety is already a client capability, not a database-name concern. Several clients declare `thread_safe = False`, including `OceanBase`, `VectorChord`, and `Adbpg`, but the fixed-rate insert runner does not use that capability. Those clients can still go through the default path and share the same DB object across insert worker threads.

Impact: streaming/fixed-rate insert benchmarks can use unsafe shared client state for some databases, causing flaky insert failures, connection/cursor errors, or inaccurate benchmark behavior. Adding a new non-thread-safe DB is also easy to get wrong because the author must know to update runner-level name branches instead of declaring the behavior in the client.

serhiizghama:
Fixes #810. The fixed-rate insert runner decided per-thread client handling by checking `db.name` against a hardcoded list (`PgVector`, `Doris`, `SeekDB`, `VolcMySQL`), so clients that declare `thread_safe = False` but aren't on that list — `OceanBase`, `VectorChord`, `Adbpg`, `LanceDB` — fell through to the default path and shared one non-thread-safe connection across insert workers. `OceanBase` was never in the list at all.

I moved the decision onto the existing `thread_safe` capability. The runner now branches on `db.thread_safe` and asks the client for a per-thread copy via a new `VectorDB.copy_for_thread()`. The default deep-copies the instance, which already works for the psycopg-family clients and for `LanceDB` (it has a custom `__deepcopy__`); the `mysql.connector`-backed clients override it to shallow-copy and drop their open socket so `init()` reconnects inside the worker. Behavior for the databases that were already special-cased is unchanged — the capability just also covers the ones the name list missed, and adding a new non-thread-safe client no longer means remembering to edit the runner.

Added a unit test that drives `send_insert_task` with a fake client and checks that thread-safe clients insert through the shared object while non-thread-safe ones go through a copy plus `init()`.


vectordb_bench/backend/clients/api.py
```diff
@@ -199,6 +200,17 @@ def prepare_filter(self, filters: Filter):
...
+    def copy_for_thread(self) -> "VectorDB":
+        """Return a per-thread copy of this client for non-thread-safe backends.
+
+        Runners call this (instead of branching on db.name) when thread_safe is
+        False, then init() the copy inside the worker so each thread owns its own
+        connection. Defaults to a deep copy; clients whose live connection can't be
+        deep-copied (e.g. an open DB-API socket) override this to shallow-copy and
+        drop their connection handles so init() re-establishes them per thread.
+        """
+        return deepcopy(self)
```
jamesgao-jpg, line 212: [P1] Avoid deep-copying live Psycopg handles

Adbpg and VectorChord both declare thread_safe=False but do not override copy_for_thread. run_with_rate keeps the parent init context open while these workers run, so the default deepcopy(self) receives a live psycopg.Connection. I reproduced this with Psycopg 3.3.4: deepcopy(connection) raises TypeError: no default __reduce__ due to non-trivial __cinit__. As a result, the first fixed-rate insert task for either backend fails before the copy reaches init().

Please give both clients a shallow-copy override that clears conn and cursor before worker init reopens them; PgVector uses the same handle pattern and is worth covering too. The fake regression client only contains copyable lists, so it does not exercise this boundary.

tests/test_rate_runner_thread_safety.py
```diff
@@ -0,0 +1,85 @@
...
+        self.thread_copy: CapturingNonThreadSafeDB | None = None
+
+    def copy_for_thread(self) -> "VectorDB":
+        c = deepcopy(self)
+        self.thread_copy = c
+        return c
+
+
+def _runner(db: VectorDB):
+    return RatedMultiThreadingInsertRunner(rate=10, db=db, dataset_iter=None)
```
jamesgao-jpg: [P2] Construct this runner with a valid current batch size

Current main defaults the insert batch size to 100 and rejects rate % batch_size != 0. This helper passes rate=10 without batch_size, so in the current merge result every helper-using test raises ValueError before reaching send_insert_task.

Please use a valid rate and batch-size pair here. The green PR workflow does not catch this because make unittest currently runs only tests/test_dataset.py::TestDataSet::test_download_small, not this new test file.

serhiizghama:
Both fixed. Gave PgVector, Adbpg and VectorChord their own shallow copy_for_thread (same pattern as Doris/OceanBase/SeekDB) so init() reconnects instead of deepcopy choking on the live psycopg handle. Also rebased on main to pick up the insert-batch-size work and fixed the test helper's rate/batch_size pair — you were right that it was raising before reaching send_insert_task.

PR #852, opened 2026-08-15
fix: include exact in MongoDB search_param

daviddallakyan2005:
## Summary

MongoDB search crashes on every query with `KeyError: 'exact'`. `search_embedding` indexes `search_params["exact"]`, but `MongoDBIndexConfig.search_param()` only returned `num_candidates_ratio`.

This adds `exact: bool = False` and emits it from `search_param()`, and reads the key with `.get("exact")`. The default stays ANN (`numCandidates`). `exact=True` enables Atlas `$vectorSearch` ENN.

The MongoDB performance UI now has a Boolean Exact (ENN) input, default False, so frontend task generation can select ENN.

## How to test

CI `make unittest` is a single network download and will not run this file.

```bash
pip install -e '.[test]'
make lint
PYTHONPATH=. python3 -m pytest tests/test_mongodb_config.py tests/test_db_client_resolution.py -q
```

### DIFF
--- a/vectordb_bench/backend/clients/mongodb/config.py
+++ b/vectordb_bench/backend/clients/mongodb/config.py
@@ -27,6 +27,7 @@ class MongoDBIndexConfig(BaseModel, DBCaseConfig):
     metric_type: MetricType = MetricType.COSINE
     num_candidates_ratio: int = 10  # Default numCandidates ratio for vector search
     quantization: QuantizationType = QuantizationType.NONE  # Quantization type if applicable
+    exact: bool = False  # Atlas $vectorSearch exact (ENN); default ANN
 
     def parse_metric(self) -> str:
         if self.metric_type == MetricType.L2:
@@ -50,4 +51,7 @@ class MongoDBIndexConfig(BaseModel, DBCaseConfig):
         }
 
     def search_param(self) -> dict:
-        return {"num_candidates_ratio": self.num_candidates_ratio}
+        return {
+            "num_candidates_ratio": self.num_candidates_ratio,
+            "exact": self.exact,
+        }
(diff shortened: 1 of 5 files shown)


vectordb_bench/backend/clients/mongodb/config.py
```diff
@@ -27,6 +27,7 @@ class MongoDBIndexConfig(BaseModel, DBCaseConfig):
     metric_type: MetricType = MetricType.COSINE
     num_candidates_ratio: int = 10  # Default numCandidates ratio for vector search
     quantization: QuantizationType = QuantizationType.NONE  # Quantization type if applicable
+    exact: bool = False  # Atlas $vectorSearch exact (ENN); default ANN
```
jamesgao-jpg, line 30: [P2] Expose exact search in the MongoDB performance UI

This adds `exact` as a public case option, but `MongoDBPerformanceConfig` still exposes only quantization and `num_candidates_ratio`. The frontend task builder instantiates `MongoDBIndexConfig` from those UI-selected fields, so every frontend-created MongoDB run keeps the `False` default and users cannot select the ENN workload described here.

Please add a Boolean `exact` input with default `False` to the MongoDB performance configuration and cover the task-generation path. If ENN is intentionally programmatic-only, please document that scope instead.

jamesgao-jpg (commented):
Thank for the contribution! Please take a look at below reviews when you get a chance. Thanks!

vectordb_bench/backend/clients/mongodb/config.py
daviddallakyan2005: I added a Boolean Exact (ENN) input, default False, to MongoDBPerformanceConfig. generate_tasks now forwards it onto MongoDBIndexConfig.exact.

PR #853, opened 2026-08-15
fix: wire MongoDB label filters through prepare_filter

daviddallakyan2005:
## Summary

MongoDB label-filter cases were silently unfiltered (recall ~0.111 vs ~0.982 once labels actually reach `$vectorSearch`). The runner already passes `labels_data=` into insert and calls `prepare_filter` then `search_embedding(query, k)`. It never passes `filters=` into search. This wires the four missing pieces, matching `qdrant_cloud`:

- declare `supported_filter_types = [NonFilter, StrEqual]` so label cases are not skipped
- persist `labels_data` on documents as `label` when `with_scalar_labels`
- `prepare_filter` stashes `{label: <value>}` for search to apply
- add `{type: "filter", path: "label"}` to the vector search index when `with_scalar_labels` (Atlas rejects undeclared filter paths)

Does not add ingest-once `metadata_fields`, dict `labels_data`, or `NumGE`.

Rebased onto #852 so production `MongoDBIndexConfig.search_param()` includes `exact`. Filter tests construct `MongoDBIndexConfig()` rather than a stub subclass.

## How to test

```bash
PYTHONPATH=. python3 -m pytest tests/test_mongodb_client.py tests/test_mongodb_config.py -q
make lint
```

`make unittest` is the network dataset download only and does not run these files.

21 passed. `make lint` clean. `tests/test_db_client_resolution.py` 43 passed.

### DIFF
--- a/tests/test_mongodb_client.py
+++ b/tests/test_mongodb_client.py
@@ -138,3 +138,91 @@ def test_create_index_omits_label_filter_without_scalar_labels():
     collection.list_search_indexes.side_effect = _ready_search_indexes()
     client._create_index()
     assert not any(field.get("type") == "filter" for field in client.index_params["fields"])
+
+
+def test_filter_supported_matches_assembler_contract():
+    assert MongoDB.filter_supported(NonFilter()) is True
+    assert MongoDB.filter_supported(LabelFilter(label_percentage=0.25)) is True
+    assert MongoDB.filter_supported(IntFilter(int_value=500, filter_rate=0.99)) is False
+
+
+def test_insert_embeddings_accepts_runner_kwargs():
+    client, collection = _make_client(with_scalar_labels=True)
+    insert_kwargs = {
+        "embeddings": [[0.1, 0.2, 0.3, 0.4]],
+        "metadata": [7],
+        "labels_data": ["label_5p"],
+    }
+    count, err = client.insert_embeddings(**insert_kwargs)
+    assert err is None
+    assert count == 1
+    documents, kwargs = collection.insert_many.call_args[0][0], collection.insert_many.call_args[1]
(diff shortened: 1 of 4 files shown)


tests/test_mongodb_client.py
```diff
@@ -0,0 +1,228 @@
...
+from vectordb_bench.backend.clients.mongodb.mongodb import MongoDB
+from vectordb_bench.backend.filter import FilterOp, IntFilter, LabelFilter, NonFilter
+
+_DB_CONFIG = {
+    "connection_string": "mongodb://localhost:27017",
+    "database": "vdb_bench",
+}
+
+
+class _IndexConfig(MongoDBIndexConfig):
```
jamesgao-jpg: [P2] Test the production config instead of hiding the PR #852 dependency

`_IndexConfig` adds `exact=False`, but the production `MongoDBIndexConfig.search_param()` on this head does not return `exact`, while `search_embedding()` reads `search_params["exact"]`. A focused reproduction using `MongoDBIndexConfig()` raises `KeyError: 'exact'` before `$vectorSearch` is built. Thus this PR only works if [PR #852](https://github.com/zilliztech/VectorDBBench/pull/852) lands first, but these tests mask that dependency. Please rebase onto #852 (and link it in the PR description), or include its fix here, then remove this subclass so the regression tests exercise the production config.

daviddallakyan2005: I rebased this onto #852 and dropped `_IndexConfig`. The filter tests now construct `MongoDBIndexConfig()`.

jamesgao-jpg:
Hi, 852 is merged and seems that a conflict needs to be resolved before 853 can get merged

PR #857, opened 2026-08-20
feat(lakebase): add Lakebase Vector backend

junyulabs:
## Summary

This PR adds [Lakebase Vector](https://neon.com/docs/extensions/lakebase-vector) backend to VectorDBBench.

## Usage

To run the `Performance768D100M` workload on [Neon](https://neon.com/):

1. Create a Neon instance with 40 CUs and wait until it is ready.
2. Enable the Lakebase Vector extension by https://neon.com/docs/extensions/lakebase-vector
3. Install and run `VectorDBBench`

```bash
export POSTGRES_HOST="<POSTGRES_HOST>"
export POSTGRES_PASSWORD="<POSTGRES_PASSWORD>"
export NUM_PER_BATCH=10000

vectordbbench lakebaseann \
  --host $POSTGRES_HOST \
  --user-name neondb_owner \
  --db-name neondb \
  --table-name vdbbench_laion100m \
  --db-label lakebase-search-laion100m \
  --task-label lakebase-search-laion100m \
  --case-type Performance768D100M \
  --k 100 \
  --load-concurrency 1 \
  --num-concurrency 1,5,10,20,30,40,60,80,100,120 \
  --concurrency-duration 30 \
  --concurrency-timeout 3600 \
  --max-parallel-workers 95 \
  --drop-old \
  --load \
  --search-concurrent \
  --search-serial \
  --probes 54,380 \
  --epsilon 3.2
```


vectordb_bench/backend/clients/lakebase_vector/lakebase_vector.py
```diff
@@ -0,0 +1,316 @@
...
+        self._set_parallel_index_build_param()
+        index_create_sql = sql.SQL(
+            "CREATE INDEX IF NOT EXISTS {index_name} ON public.{table_name} "
+            "USING {index_type} ({vector_field} {metric})"
+        ).format(
+            index_name=sql.Identifier(self._index_name),
+            table_name=sql.Identifier(self.table_name),
+            index_type=sql.SQL(index_param["index_type"]),
+            vector_field=sql.Identifier(self._vector_field),
+            metric=sql.SQL(index_param["metric"]),
```
yhmo, line 203: vectordb_bench/backend/clients/lakebase_vector/lakebase_vector.py line:203
High ---- `metric_type` is never set for this backend: `cli.py` hardcodes `metric_type=None` (line 74), the frontend case-config lists in `dbCaseConfigs.py` expose no metric input, and the task runner passes `db_case_config` through unchanged. `LakebaseANNConfig.parse_metric()` therefore returns `None`, so `index_param()["metric"]` is `None` and `sql.SQL(None)` raises `TypeError` ("SQL values must be strings") inside `_create_index()` during `optimize()`. That breaks the exact CLI workflow documented in the PR body (load succeeds, optimize crashes before search). The sibling VectorChord client handles this by defaulting to cosine ops (`ops.get(self.metric_type, ops[MetricType.COSINE])`); if you copy that default here, note `parse_metric_fun_op()` already falls back to `<=>` (cosine) for `None`, which would still mismatch recall computed against the dataset's real metric (e.g. L2 for LAION). Please either wire a metric_type input into both CLI and frontend, or derive it from the dataset metric, and add a config-level test asserting `index_param()["metric"]` is a valid operator class.

vectordb_bench/backend/clients/lakebase_vector/config.py
```diff
@@ -0,0 +1,105 @@
...
+                "dbname": self.db_name,
+                "user": user_str,
+                "password": pwd_str,
+            },
+            "table_name": self.table_name,
+        }
+
+
+class LakebaseANNConfig(BaseModel, DBCaseConfig):
+    metric_type: MetricType | None = None
```
yhmo, line 46: vectordb_bench/backend/clients/lakebase_vector/config.py line:46
Medium ---- This new backend ships without any tests, while the repo has clear precedents for PG clients (e.g. `tests/test_pgvector.py` runs an e2e against a dockerized pgvector, and `tests/test_milvus_zilliz_cli.py` covers CLI registration). None of the material behaviors — metric/operator-class derivation, index create/drop after load, session GUC (probes/epsilon) rendering, filtered search — have direct test evidence. A minimal config test (e.g. assert `LakebaseANNConfig(...).index_param()` returns a usable metric and that the composed CREATE INDEX is well-formed) would have caught the TypeError in the metric path above.

vectordb_bench/backend/clients/lakebase_vector/lakebase_vector.py
```diff
@@ -0,0 +1,316 @@
...
+        self._scalar_label_field = "label"
+
+        # construct basic units
+        self.conn, self.cursor = self._create_connection(**self.connect_config)
+
+        # create lakebase_vector extension
+        self.cursor.execute("CREATE EXTENSION IF NOT EXISTS lakebase_vector CASCADE")
+        self.conn.commit()
+
+        log.info(f"{self.name} config values: {self.connect_config}\n{self.case_config}")
```
yhmo: vectordb_bench/backend/clients/lakebase_vector/lakebase_vector.py line:63
Low ---- `self.connect_config` includes the plaintext `password` (set in `LakebaseVectorConfig.to_dict()`), and this INFO log dumps it at every client construction. The existing pgvector/vectorchord clients do the same, so this matches repo convention, but it is still a real credential exposure in logs. Consider logging a redacted copy (mask password) when touching these files.

vectordb_bench/backend/clients/lakebase_vector/cli.py
```diff
@@ -0,0 +1,80 @@
...
+        click.option(
+            "--password",
+            type=str,
+            default=lambda: os.environ.get("POSTGRES_PASSWORD", ""),
+            show_default="$POSTGRES_PASSWORD",
+        ),
+    ]
+    host: Annotated[str, click.option("--host", type=str, required=True)]
+    port: Annotated[int, click.option("--port", type=int, default=5432, show_default=True)]
+    db_name: Annotated[str, click.option("--db-name", type=str, default="databricks_postgres", show_default=True)]
```
yhmo: vectordb_bench/backend/clients/lakebase_vector/cli.py line:25
Low ---- The default `--db-name databricks_postgres` (also `db_name` in `LakebaseVectorConfig`) looks like a copy-paste leftover: it has nothing to do with Lakebase/Neon and would silently target a nonexistent database when the flag is omitted. Use a neutral default (e.g. `postgres`) or make the flag required like `--host`.

vectordb_bench/backend/clients/lakebase_vector/config.py
junyulabs: Thanks for raising this. Now we have added the test file for the `lakebase_vector` backend.

vectordb_bench/backend/clients/lakebase_vector/lakebase_vector.py
junyulabs: Thanks for pointing this out. We initially followed the PGVector implementation, but agree that logging `connect_config` exposes credentials. We have removed it from the log and now only output the case configuration.

vectordb_bench/backend/clients/lakebase_vector/lakebase_vector.py
junyulabs: In the current workflow, `Assembler.assemble_all()` calls `Assembler.assemble()`, which sets `task.db_case_config.metric_type = c.dataset.data.metric_type` before creating the client, as it does for PGVector. So this cannot happen.

As a defensive measure, both metric parsing methods now raise `NotImplementedError` when the metric is absent or unsupported, preventing None from reaching SQL construction or silently falling back to cosine.

Tests verify that Assembler derives the metric from L2 and cosine datasets, and that L2, IP, and cosine produce valid operator classes, search operators, and complete CREATE INDEX SQL.

vectordb_bench/backend/clients/lakebase_vector/cli.py
junyulabs: Fixed. Now it is a required flag.

vectordb_bench/backend/clients/lakebase_vector/config.py
```diff
@@ -0,0 +1,107 @@
...
+        return {"metric_fun_op": self.parse_metric_fun_op()}
+
+    def session_param(self) -> LakebaseSessionCommands:
+        session_options = []
+        if self.probes is not None and self.probes.strip():
+            session_options.append(
+                {
+                    "parameter": {
+                        "setting_name": "lakebase_ann.probes",
+                        "val": self.probes,
```
yhmo, line 89: vectordb_bench/backend/clients/lakebase_vector/config.py line:89
Medium ---- The --probes value is forwarded verbatim (e.g. "54,380" in the PR usage example, or "10,20" in test_session_guc) into a single SET "lakebase_ann.probes" = "54,380";. The linked lakebase_vector documentation (neon.com/docs/extensions/lakebase-vector) declares this GUC as an integer ("Number of IVF partitions to scan at query time", example SET lakebase_ann.probes TO '10'). If the server parses it as a plain integer, every init()/session would fail before any search runs. The CLI/frontend help advertises "Comma-separated lakebase_ann probe counts", yet nothing in the code splits or iterates that list - only one SET is emitted. Could you confirm whether the server actually accepts a comma-separated value and, if so, how the single value is meant to apply across the concurrent search runs? If multiple probe values were intended, the framework has no expansion for them; otherwise an int with client-side validation would match the documented contract. It would also help to cover this in a test that asserts what the server accepts, not just the rendered SET SQL.

vectordb_bench/backend/clients/lakebase_vector/lakebase_vector.py
```diff
@@ -0,0 +1,316 @@
...
+                        copy.write_row((row, embeddings_arr[i], labels_data[i]))
+                    else:
+                        copy.set_types(["bigint", "vector"])
+                        copy.write_row((row, embeddings_arr[i]))
+            self.conn.commit()
+
+            return len(metadata), None
+        except Exception as e:
+            log.warning(f"Failed to insert data into lakebase_vector table ({self.table_name}), error: {e}")
+            return 0, e
```
yhmo, line 280: vectordb_bench/backend/clients/lakebase_vector/lakebase_vector.py line:279
Medium ---- On a COPY failure this returns (0, e) without rolling back the connection. Since _create_connection sets autocommit=False, a failed COPY statement marks the whole transaction aborted; the load runner then retries the batch up to MAX_INSERT_RETRY (5) times on that same connection (ConcurrentInsertRunner._insert_batch_with_retry), and every retry fails with "current transaction is aborted, commands ignored until end of transaction block", so one bad batch fails the entire load and masks the original error. A single self.conn.rollback() in this except handler would keep the connection retryable, and a test asserting insert_embeddings returns (0, e) then a subsequent call succeeds would lock the behavior in.

vectordb_bench/backend/clients/lakebase_vector/lakebase_vector.py
```diff
@@ -0,0 +1,316 @@
...
+            metadata_arr = np.array(metadata)
+            embeddings_arr = np.array(embeddings)
+            with self.cursor.copy(
+                sql.SQL("COPY public.{table_name} FROM STDIN (FORMAT BINARY)").format(
+                    table_name=sql.Identifier(self.table_name),
+                )
+            ) as copy:
+                for i, row in enumerate(metadata_arr):
+                    if self.with_scalar_labels:
+                        copy.set_types(["bigint", "vector", "varchar"])
```
yhmo: vectordb_bench/backend/clients/lakebase_vector/lakebase_vector.py line:269
Low ---- copy.set_types(...) is re-invoked for every row inside the loop. psycopg Copy.set_types() re-registers the per-column dumpers via transformer.set_dumper_types() on each call, so for the 100M-row Performance768D100M load this is avoidable per-row work in the hottest copy path. The type set only depends on with_scalar_labels, so a single set_types(...) call before the loop would be equivalent. Note test_insert currently asserts set_types.call_count == 2 for two rows, which locks in the per-row behavior and would need updating if the call is hoisted.

vectordb_bench/backend/clients/lakebase_vector/config.py
```diff
@@ -0,0 +1,107 @@
...
+class LakebaseSessionCommands(TypedDict):
+    session_options: Sequence[dict[str, Any]]
+
+
+class LakebaseVectorConfig(DBConfig):
+    user_name: SecretStr = SecretStr("postgres")
+    password: SecretStr
+    host: str = "localhost"
+    port: int = 5432
+    db_name: str = "databricks_postgres"
```
yhmo: vectordb_bench/backend/clients/lakebase_vector/config.py line:27
Low ---- The earlier thread on cli.py --db-name was resolved by making the flag required, but this config-class default db_name: str = "databricks_postgres" was left in place. It is unrelated to Lakebase/Neon, and the non-CLI path still uses it: dbConfigSetting builds the frontend form from the model schema default, so a frontend user who leaves db_name untouched will silently target a nonexistent database and the run fails at connect time. Suggest a neutral default (e.g. "postgres") or no default, consistent with vectorchord ("vectordb") and alloydb (required).

junyulabs: Fixed by making `db_name` required like AlloyDB、Hologres、PgDiskANN、PgVecto_rs、PgVectorScale.

vectordb_bench/backend/clients/lakebase_vector/lakebase_vector.py
junyulabs: Thanks. I moved `copy.set_types()` outside the row loop. and updated the test

vectordb_bench/backend/clients/lakebase_vector/lakebase_vector.py
junyulabs: `self.conn.rollback()` is added. I think the `pgvector` backend might has the same issue.

vectordb_bench/backend/clients/lakebase_vector/config.py
junyulabs: The `lakebase_vector` accepts either a single positive integer for single-level IVF (for example, 10) or comma-separated positive integers for multi-level IVF (for example, 54,380). 

We clarified this format in the CLI and frontend help text and parameterized the existing session GUC test to cover both single-level and two-level IVF configurations.

junyulabs:
Hi @yhmo, the previous review comments have been addressed. 

When you have a chance, could you please take another look? Please let me know if any further changes are needed. Thank you!

junyulabs:
Hi @yhmo, gentle follow-up on this PR. All previous review comments have been addressed and the related threads are resolved. When you have a chance, could you please take another look?

vectordb_bench/backend/clients/lakebase_vector/lakebase_vector.py
```diff
@@ -0,0 +1,317 @@
...
+from ..api import VectorDB
+from .config import LakebaseANNConfig, LakebaseVectorConfigDict
+
+log = logging.getLogger(__name__)
+
+
+class LakebaseVector(VectorDB):
+    """Use psycopg instructions"""
+
+    thread_safe: bool = False
```
jamesgao-jpg, line 25: [P2] Apply #849's connection-safe thread-copy override to Lakebase

VERIFIED: This client sets `thread_safe = False` but inherits `VectorDB.copy_for_thread()`, which uses `deepcopy(self)`. The streaming insert runner opens the parent connection before dispatching workers, so copying this client attempts to deep-copy its live psycopg connection and raises `TypeError` before insertion. The offline reproduction with a real psycopg connection object failed with `no default __reduce__ due to non-trivial __cinit__`.

https://github.com/zilliztech/VectorDBBench/pull/849 already fixed this for the existing psycopg clients, and that commit is already included in this branch; rebasing alone will not fix the new client. Please add the same override as [PgVector](https://github.com/zilliztech/VectorDBBench/blob/b6fd49d52e5d3fe57163671b0569e5c7d9465f6d/vectordb_bench/backend/clients/pgvector/pgvector.py#L103-L109): shallow-copy the client, clear `conn`/`cursor`, and let the worker's `init()` reconnect. Add a regression test covering a parent with a non-deepcopyable connection.

jamesgao-jpg:
Hi, thanks for the contribution! I've posted a rather delayed review and once that's resolved I will consult @yhmo and get it merged soon. Thanks again!

vectordb_bench/backend/clients/lakebase_vector/lakebase_vector.py
junyulabs: Thanks for review. I have applied the #849 to the backend, together with a new regression test `test_copy_for_thread`.

PR #858, opened 2026-08-24
feat(valkey): add Valkey backend

bluayer:
## Summary

Adds a Valkey backend using `valkey-glide-sync`.

- HNSW vector search
- numeric and label filters
- cluster and cluster-mode-disabled connections with TLS
- configurable request and connection timeouts
- pipelined inserts and prefix-scoped cleanup


vectordb_bench/backend/clients/valkey/valkey.py
```diff
@@ -0,0 +1,266 @@
...
+
+    def _delete_prefixed_documents(self, conn: ValkeyConnection) -> None:
+        deleted = 0
+        pending = 0
+        batch = self._new_batch()
+        for key in self._scan_keys(conn):
+            batch.unlink([key])
+            deleted += 1
+            pending += 1
+            if pending == config.NUM_PER_BATCH:
```
yhmo: vectordb_bench/backend/clients/valkey/valkey.py line:135
Medium ---- drop_old cleanup flushes UNLINK in batches of config.NUM_PER_BATCH (default 100) and SCANs with count=100, so re-running a case over an existing dataset costs a large number of round trips (roughly 20k+ exec/scans for 1M keys, and over 1M for the 100M-vector cases) before the timed load even starts. Consider a dedicated, larger scan-count/flush constant, or issuing FT.DROPINDEX <name> DD via custom_command when all prefix documents are known to be indexed (glide's ft.dropindex has no DD option at v2.5.1, so a raw command is needed).

tests/test_valkey.py
```diff
@@ -0,0 +1,291 @@
+from typing import Any
+from unittest.mock import patch
+
+import numpy as np
+import pytest
+from glide_sync import Batch
```
yhmo, line 6: tests/test_valkey.py line:6
Low ---- This module-level import of glide_sync is only satisfied by the [valkey] extra, which CI does not install (pull_request.yml installs ".[test]" only), so these new tests cannot be collected in the standard test environment and are not exercised by `make unittest` or `make lint`. Consider pytest.importorskip("glide_sync") (or adding the valkey extra to the [test] dependency set) and wiring the tests into the CI test command so this substantial coverage can actually run.

vectordb_bench/backend/clients/valkey/config.py
```diff
@@ -0,0 +1,58 @@
+from pydantic import BaseModel, Field, SecretStr
+
+from ..api import DBCaseConfig, DBConfig, IndexType, MetricType
+
+
+class ValkeyConfig(DBConfig):
+    password: SecretStr | None = None
+    host: SecretStr
+    port: int = Field(default=6379, ge=1, le=65535)
+    ssl: bool = True
```
yhmo, line 10: vectordb_bench/backend/clients/valkey/config.py line:10
Low ---- TLS is enabled by default (ssl=True) but there is no CA-certificate path or insecure-TLS toggle, even though glide supports both (tls_config / use_insecure_tls in AdvancedGlide*ClientConfiguration) and the sibling Redis/MemoryDB backends expose ssl_ca_certs. Users with private-CA TLS Valkey cannot connect, and plain local Valkey setups silently fail until they discover --no-ssl. Consider adding ssl_ca_certs (or an insecure-tls flag) and wiring it through _create_client.

vectordb_bench/backend/clients/valkey/valkey.py
bluayer: Valkey Search does not support DD, so I kept FT.DROPINDEX and changed SCAN/UNLINK cleanup to use a dedicated 10K key chunk.

vectordb_bench/backend/clients/valkey/config.py
bluayer: Added --insecure-tls for both client modes and rejected its use with --no-ssl.

tests/test_valkey.py
bluayer: @yhmo Optional backend tests are not currently part of the standard PR CI, so Valkey backend also follows the existing convention. Is this a new requirement for all new backends?

yhmo: Thanks for the clarification. I agree that no change is needed here.

bluayer:
I have addressed all the points raised in the review. Please let me know if there are any further steps needed to move this PR forward.

vectordb_bench/backend/clients/valkey/valkey.py
```diff
@@ -0,0 +1,272 @@
...
+            yield
+        finally:
+            self.conn = None
+            conn.close()
+
+    def optimize(self, data_size: int | None = None) -> None:
+        """Valkey does not require a post-load optimization step."""
+        return
+
+    def insert_embeddings(
```
baswanth09: @bluayer You can also add an option to make this multi-threaded on the client side. This will improve the ingestion time. Basically you would want the valkey main thread to be saturated
For example: https://github.com/zilliztech/VectorDBBench/commit/233eebe91327ab06a03083fa356d71a205854a6b

PR #860, opened 2026-08-24
feat: add support for PolarDB for PostgreSQL HNSW

cxyybless:
## Summary

This PR adds a pgvector-compatible PolarDB for PostgreSQL HNSW client
to VectorDBBench. It is separate from the existing MySQL-compatible
PolarDB client and does not change the existing `polardb*` commands.

## Supported features

- HNSW index creation and search
- PQ, SQ4, SQ8, and RaBitQ internal quantization
- CLI command: `vectordbbench polardbpghnsw`
- Streamlit configuration and result visualization
- Optional Graph Cache readiness handling before search benchmarks
- Reuse of the existing `pgvector` dependency extra

PolarDB-specific internal quantization and Graph Cache options require
PolarDB vector extension 0.8.3.1 or later.

## Implementation

- Reuse the PgVector client and HNSW configuration where behavior is
  compatible.
- Add PolarDB-specific index options and validation without duplicating
  the PostgreSQL client implementation.
- Build or validate Graph Cache after index creation and for search-only
  runs when requested.
- Redact PostgreSQL connection passwords from client logs.
- Add PolarDBPG configuration, lifecycle, CLI, and frontend tests.

## Testing

- Python 3.11.15:
  - `make lint`
  - existing dataset unit test
  - 76 PolarDBPG and client-resolution tests
- Python 3.12.13:
  - `make lint`
  - existing dataset unit test
  - 76 PolarDBPG and client-resolution tests
  - wheel build with `pip wheel --no-deps .`
- PolarDB for PostgreSQL 16.14 with vector extension 0.8.3.1:
  - full 50K x 1536D load and RaBitQ 8-bit HNSW index build
  - Graph Cache reached `ready` and `usable=yes` before search
  - Recall@100: 0.9909
  - search-only runs with Graph Cache lifecycle enabled and skipped
  - plain, PQ, SQ4, and SQ8 HNSW index creation and query smoke tests
  - verified that logs and result files do not expose the database password


vectordb_bench/backend/clients/polardb_pg/polardb_pg.py
```diff
@@ -0,0 +1,176 @@
...
+
+class PolarDBPgHNSW(PgVector):
+    """PgVector-compatible PolarDB HNSW client."""
+
+    case_config: PolarDBPgHNSWConfig
+
+    def __init__(self, *args, drop_old: bool = False, **kwargs):
+        super().__init__(*args, drop_old=drop_old, **kwargs)
+        self.name = "PolarDBPG"
+
```
yhmo, line 68: vectordb_bench/backend/clients/polardb_pg/polardb_pg.py line:69
Low ---- self.name is assigned after super().__init__() returns, so every log line emitted during PgVector.__init__ (config values, create/drop table, create/drop index) and the "config must create an index" RuntimeError message still print "PgVector" for PolarDBPG runs. Consider setting self.name before delegating to the base constructor (or overriding it as a class attribute) so operators can filter PolarDBPG logs consistently.

vectordb_bench/models.py
```diff
@@ -173,6 +173,13 @@ class CaseConfigParamType(Enum):
     post_load_index = "post_load_index"
     pq_nbits = "pq_nbits"
 
+    # PolarDB for PostgreSQL parameters
+    hnsw_quantization = "quantization"
```
yhmo: vectordb_bench/models.py line:177
Low ---- hnsw_quantization = "quantization" duplicates the existing mongodb_quantization_type = "quantization" value, so Python's Enum makes hnsw_quantization a silent alias (CaseConfigParamType.hnsw_quantization.name == "mongodb_quantization_type", and the member is skipped when iterating the enum). The docstring says the value is the key used in CaseConfig.params and the UI, so two unrelated DBs now share the key "quantization". Nothing currently does a reverse lookup, but any future CaseConfigParamType(key) or name-based serialization cannot distinguish them. Consider a distinct value such as "hnsw_quantization".

tests/test_polardb_pg.py
```diff
@@ -0,0 +1,357 @@
...
+            1.0,
+        )
+        monkeypatch.setattr("vectordb_bench.backend.clients.polardb_pg.polardb_pg.time.monotonic", monotonic)
+
+        with pytest.raises(TimeoutError, match="status=building"):
+            client._ensure_graph_cache_ready()
+
+
+def test_polardb_pg_cli_help():
+    result = CliRunner().invoke(PolarDBPgHNSWCommand, ["--help"])
```
yhmo, line 348: tests/test_polardb_pg.py line:306
Medium ---- The CLI is only exercised via --help, which never builds PolarDBPgConfig/PolarDBPgHNSWConfig or reaches run(); the ~30-option plumbing (PgVectorTypedDict + HNSWFlavor1 + PolarDBPgHNSWOptions -> config constructors -> TaskConfig) is therefore untested. A test invoking the command with required options plus --dry-run would validate that wiring without a live database, since run() returns before benchmark_runner.run() when dry_run is set.

vectordb_bench/backend/clients/polardb_pg/polardb_pg.py
```diff
@@ -0,0 +1,176 @@
...
+        # Search-only benchmarks skip optimize(), so validate and prepare the
+        # existing index before any search workers can start.
+        if not drop_old and self.case_config.graph_cache:
+            with self.init():
```
yhmo, line 72: vectordb_bench/backend/clients/polardb_pg/polardb_pg.py line:73
Low ---- For search-only runs (drop_old=False) the graph-cache readiness wait runs synchronously inside __init__ during _pre_run, outside the case-level optimize/load timeout wrappers, and is capped only by the hard-coded _GRAPH_CACHE_TIMEOUT_SECONDS = 3600 with no CLI/frontend knob (the tests explicitly assert --graph-cache-timeout is absent). A stuck or slow cache build therefore holds the runner process for up to an hour with no way to tune it. Consider exposing the timeout (or poll interval) as an option so operators can bound search-only runs.

vectordb_bench/backend/clients/polardb_pg/polardb_pg.py
cxyybless: Fixed in 7b94d1c. `PgVector.name` is now a class attribute and `PolarDBPgHNSW` overrides it with `PolarDBPG`, so logs emitted by the base constructor use the correct client name. Added regression coverage for the initialization log.

vectordb_bench/models.py
cxyybless: Fixed in 7b94d1c. `hnsw_quantization` now has a distinct enum value. `PolarDBPgHNSWConfig` accepts both the frontend `hnsw_quantization` key and the CLI `quantization` key, preserving index-option generation. Added coverage for enum identity and frontend-to-config mapping.

tests/test_polardb_pg.py
cxyybless: Fixed in 7b94d1c. Added a real `polardbpghnsw --dry-run` test with Graph Cache and RaBitQ options. It verifies TaskConfig construction, option propagation, skipped benchmark execution, and password redaction.

vectordb_bench/backend/clients/polardb_pg/polardb_pg.py
cxyybless: Fixed in 7b94d1c. Graph Cache wait timeout is now configurable through the Pydantic config, CLI `--graph-cache-timeout`, and Streamlit UI, with positive-integer validation and a 3600-second default. Lifecycle, CLI, and frontend tests cover the new setting.

cxyybless:
@yhmo The four review comments have been addressed in d41c850. The branch is rebased onto the current main and remains a single commit. Focused tests pass (82 passed), and Black/Ruff pass. Could you please take another look when convenient?

@xuanyang-cn Could you please approve the pending GitHub Actions workflow?

PR #867, opened 2026-08-31
fix: make QdrantLocal runnable from the web UI

### LINKED ISSUE #796: [BUG]: QdrantLocal Error
I am trytin to benchmark a local qdrant vector database. I run into error BEFORE starting a test

Image

Image

Image

Seems like 'ef_construct' is missing, but there is not such option

serhiizghama:
Fixes #796. QdrantLocal can't be run from the web UI at all — `QdrantLocalIndexConfig` requires `m` and `ef_construct`, but there's no `DB.QdrantLocal` entry in `CASE_CONFIG_MAP`, so the UI renders "Auto", passes an empty config, and pydantic raises `2 validation errors ... Field required` before the run starts. That matches the reporter's "seems like 'ef_construct' is missing, but there is not such option" — the option genuinely isn't there.

Gave both fields the same defaults the CLI already uses (`m=16`, `ef_construct=200`) and registered QdrantLocal in `CASE_CONFIG_MAP` so `m`, `ef_construct` and `on_disk` are tunable for load, plus `hnsw_ef` for search — the same four knobs `qdrant_local/cli.py` exposes. Needed two new `CaseConfigParamType` members since the existing `ef_construction` spelling doesn't match the model field.

The CLI path is unchanged: it always passed these explicitly. Verified the empty-config case now builds, that the UI defaults round-trip into `index_param()`/`search_param()`, and that tuned values reach the client. New tests fail on main and pass here.


vectordb_bench/frontend/config/dbCaseConfigs.py
```diff
@@ -3293,6 +3293,63 @@ class FilterType(Enum):
     CaseConfigParamInput_NovaAdaptiveGamma_Adbpg,
 ]
 
+CaseConfigParamInput_m_QdrantLocal = CaseConfigInput(
+    label=CaseConfigParamType.m,
+    displayLabel="m",
+    inputHelp="HNSW index parameter m, set 0 to disable the index",
```
yhmo: vectordb_bench/frontend/config/dbCaseConfigs.py line:3299
Low ---- The new m input lets users set m=0 with the help text "set 0 to disable the index", but no such code path exists: QdrantLocalIndexConfig.index_param() always forwards m into HnswConfigDiff(m=...), and Qdrant does not validate m (only ef_construct is constrained to min=4), so create_collection accepts m=0 and builds an HNSW graph with m0 = 2*m = 0 edges — a degenerate index that silently yields broken search results rather than a disabled index. Consider setting min=2, rewording/removing the "set 0 to disable the index" claim, or implementing an actual no-index path.

serhiizghama:
You're right — that help text was inherited from the local CLI's `--m` option, and it doesn't hold here. `m=0` is Qdrant's multitenancy knob: it disables the *global* index only when `payload_m` builds per-tenant ones instead, which `qdrant_cloud` sets but `qdrant_local` never does. So in this client 0 just leaves the collection without a vector index.

Bumped the input to `min=2` and reworded the help to describe `m` plainly. Fixed the same sentence in the CLI too, so the two surfaces don't disagree about the same flag. Added a small test on the bound.

vectordb_bench/frontend/config/dbCaseConfigs.py
yhmo: The issue is confirmed fixed.

Issue #825, opened 2026-07-29
[Question/Enhancement] Unconditional force-merge compaction in Milvus client `_optimize()` may be too costly and non-reproducible at 100M scale (LAION-100M)


frankleaf:
## Summary

In the Milvus client, `_optimize()` unconditionally triggers a **force-merge compaction**
before search for all non-GPU index types:

https://github.com/zilliztech/VectorDBBench/blob/main/vectordb_bench/backend/clients/milvus/milvus.py#L312-L316

```python
MILVUS_FORCE_MERGE_TARGET_SIZE_MB = ((1 << 63) - 1) // (1024**2)   # line 21
...
self._wait_for_segments_sorted()
self._wait_for_index()
compaction_id = self.client.compact(
    self.collection_name, target_size=MILVUS_FORCE_MERGE_TARGET_SIZE_MB
)
```

Concerns at LAION-100M scale
I'd like to ask whether this is intended to run at 100M-vector scale (e.g. the
Performance768D100M / LAION-100M case), and whether it should be made configurable.

LAION-100M is 100M × 768-dim float32 ≈ 307 GB of raw vector data.
* Compaction cost: force merge rewrites the segments; at 300 GB+ this is a very long,
IO/CPU-heavy operation on DataNodes. _optimize() is supposed to be a "prepare for
search" step, but here it becomes a multi-hour data rewrite.
* Index rebuild memory: after merging into much larger segments, the vector index is
rebuilt per segment. Index construction memory scales with segment size, so very large
merged segments raise real OOM risk on the index/data node — especially for
memory-hungry builds (HNSW, DiskANN).

Questions
1、Was unconditional force merge validated on the 100M-scale cases, and if so, what was
the observed compaction duration and peak DataNode / IndexNode memory?
2、Is the intent to normalize segment layout for fair comparison? If yes, wouldn't an
explicit, fixed target_size be more reproducible than the memory-dependent auto mode?

xiaofan-luan:
for milvus, it's just tune to a reason size, doesn't mean you only merge to one segment.

we might end up with 10 30GB segments. and with more machines in your benchmark, it might be more.

if you have streaming use case you can skip the optimize stage.

frankleaf:
@xiaofan-luan 
[...]

PR #871, opened 2026-09-15
feat: add CLI control for Milvus force-merge size and toggle

jamesgao-jpg:
## What

Adds CLI control of the Milvus force-merge compaction in `optimize()`,
addressing #825 (the unconditional force merge is costly and
non-reproducible at 100M scale).

Two new options on every Milvus CLI command and on `ZillizAutoIndex`:

- `--force-merge-enabled/--no-force-merge-enabled` — skip the force-merge
  stage for a sooner ready-to-search state (segments are then not merged to
  their fullest potential).
- `--force-merge-target-size-mb` — request a bounded merged segment size
  (e.g. `1024` MB); the effective cap depends on the Milvus server. Default
  keeps the current unbounded single-segment behavior
  (`((1 << 63) - 1) // (1024**2)` MB).

Both are also exposed in the web UI case configs for Milvus and Zilliz
Cloud (Load / Performance / FTS).

## Behavior

- `_optimize()` keeps flush → normal compaction → index-wait → refresh.
  When force merge is disabled, only the force-merge stage is skipped.
- `_force_merge()` resolves the target size from the case config; `None`
  falls back to today's constant, so existing runs are byte-for-byte
  unchanged.
- GPU index types keep skipping force merge (unchanged).
- `force_merge_target_size_mb` is validated as a positive integer or `None`.

## Compatibility

- Backward compatible: defaults (`enabled=True`, `target=None`) reproduce
  current behavior; no result-schema or metric changes (optimize is
  pre-search preparation, not a measured result).
- Zilliz Cloud inherits the Milvus client and config base classes, so the
  same knobs apply there too.

## Testing

- Level 1 mock-only unit tests: `tests/test_milvus.py` +
  `tests/test_milvus_zilliz_cli.py`, 31 passed / 1 integration deselected,
  including new tests for the configured target size, the disabled path,
  config validation, and CLI flag mapping (Milvus / FTS / Zilliz Cloud).
- No live-backend probe yet: whether a bounded `target_size` is honored as
  a hard segment-size cap by the target Milvus server still needs a small
  functional probe before benchmark readiness can be declared.

## Limitations / open items

- Functional probe for bounded target-size semantics on a live Milvus.
- [WIP] — review welcome before finalizing defaults and docs.

## Follow-up fix (84beccd)

- \`copy_fts_compatible_db_case_fields()\` now also preserves
  \`force_merge_target_size_mb\` when a non-FTS vector config is routed to an
  FTS config class, so \`--force-merge-target-size-mb\` survives a
  \`--case-type FTSBm25Performance\` invocation.
- Added \`test_cli_preserves_milvus_force_merge_options_when_routing_vector_config_to_fts\`.


vectordb_bench/backend/clients/milvus/cli.py
```diff
@@ -23,8 +23,14 @@ def _use_partition_key(parameters: dict) -> bool:
     return parameters.get("case_type") == "CloudMultiTenantSearchCase"
 
 
-def _with_partition_key(db_case_config: BaseModel, parameters: dict) -> BaseModel:
-    return db_case_config.model_copy(update={"use_partition_key": _use_partition_key(parameters)})
+def _apply_milvus_case_defaults(db_case_config: BaseModel, parameters: dict) -> BaseModel:
+    return db_case_config.model_copy(
```
jamesgao-jpg, line 40: [P2] Validate the copied target size

`model_copy(update=...)` does not run Pydantic validators, so the validator in `MilvusIndexConfig` is bypassed on every regular Milvus command. VERIFIED: both `--force-merge-target-size-mb 0` and `-5` exit successfully and produce configs containing those values; PyMilvus rejects them only later in `compact()`, after a benchmark may already have loaded its dataset. Please use a CLI `IntRange(min=1)` or rebuild/revalidate the model, and add zero/negative CLI regressions.

vectordb_bench/backend/clients/milvus/cli.py
```diff
@@ -98,6 +104,35 @@ class MilvusTypedDict(TypedDict):
...
+        click.option(
+            "--force-merge-target-size-mb",
+            type=int,
+            required=False,
+            default=None,
+            show_default=True,
+            help=(
+                "Target merged segment size in MB for the force-merge compaction during "
+                "optimize. Defaults to the current unbounded single-segment behavior; set "
+                "e.g. 1024 for bounded, reproducible segments. Must be a positive integer."
```
jamesgao-jpg: [P2] Describe this as a nominal target, not a reproducibility guarantee

The public wording is stronger than the backend contract. [PyMilvus 2.6.16](https://github.com/milvus-io/pymilvus/blob/v2.6.16/pymilvus/milvus_client/optimize_task.py#L22-L24) describes its maximum-size sentinel as letting the server choose the segment size automatically, while the [Milvus force-merge docs](https://milvus.io/docs/force-merge.md) describe explicit values as approximate and safety-clamped by topology/memory. The PR also notes that no live probe established a hard cap. Since segment layout affects benchmark latency, please replace `unbounded single-segment` / `bounded, reproducible` in the CLI, UI, and PR description with `server-selected automatic target` / `requested or nominal target`, plus the backend constraints.

jamesgao-jpg (commented):
Reviewed the effective diff at `84beccd`. The force-merge toggle, target propagation, frontend wiring, GPU behavior, and FTS routing otherwise look correct. Focused Milvus/CLI/FTS tests passed (38 passed), and all four current CI jobs are green. Two findings remain below: regular Milvus CLI commands bypass the positive-size validator, and the public target-size wording overstates the backend's guarantees. No live Milvus/Zilliz functional probe was run.

jamesgao-jpg:
Both findings addressed in `72b1219` (pushed).

**1. CLI bypass of the positive-size validator — fixed.**

Confirmed the mechanism from pydantic source: `BaseModel.model_copy(update=...)` documents "the data is not validated before creating the new model" (`pydantic/main.py:405`). Since `_apply_milvus_case_defaults` applies the flag via `model_copy`, `--force-merge-target-size-mb 0` (or negative) reached the config without ever hitting the `field_validator` on the 17 index commands (the FTS command constructs directly and was already guarded).

Fix: a click callback `_validate_positive_int_or_none` on the option in both `milvus/cli.py` and `zilliz_cloud/cli.py` rejects non-positive values at the CLI boundary (`click.BadParameter`), while direct config construction remains guarded by the pydantic validator. Added `test_milvus_autoindex_cli_rejects_non_positive_force_merge_target_size` and `test_zilliz_autoindex_cli_rejects_non_positive_force_merge_target_size` (0 and -5 → non-zero exit + message).

**2. Public wording overstates backend guarantees — fixed.**

The CLI help and web-UI `inputHelp` now describe the value as a *requested* cap whose effective behavior depends on the Milvus server (no claim of guaranteed bounded/reproducible segments until the functional probe is run).

Verification: focused Milvus/CLI/FTS tests on the remote worktree at `72b1219` → **39 passed, 1 integration deselected**. CI on the updated head will re-run; the live Milvus/Zilliz functional probe remains the open item before benchmark readiness (tracked in the PR description).

PR #874, opened 2026-09-15
feat: support configurable NQ for concurrent Milvus searches

### LINKED ISSUE #873: [Feature Request] Support configurable NQ for concurrent vector search
VectorDBBench sends one query vector per search request. Changing concurrency cannot
represent applications that send batched searches.

Add `--nq` (YAML: `nq`) to set the number of query vectors per concurrent request.
Default to `1` to preserve existing behavior. Start with Milvus and Zilliz Cloud
single-tenant vector performance cases, including filtered searches.

## Expected behavior

- Keep the configured concurrency list and record NQ in results; keep different NQ values separate in result views.
- Count successful query vectors per second, excluding failed or incomplete batches; measure concurrent latency per successful batch request.
- Keep serial latency and recall at NQ=1, concurrency=1.
- Reject unsupported backends and workloads before data loading.

## Example

Search an existing collection with two query vectors per request. The collection must
contain the selected dataset and a matching HNSW index.

```shell
vectordbbench milvushnsw \
  --uri http://localhost:19530 --collection-name VDBBench \
  --case-type Performance768D1M --m 16 --ef-construction 128 --ef-search 128 \
  --k 10 --nq 2 \
  --skip-drop-old --skip-load
```

YAML uses `nq: 2`. Omitting `--num-concurrency` uses the configured concurrency list.

yanbinyang:
Fixes #873. Concurrent searches currently send one query vector per request and do not expose a per-request query batch size. VectorDBBench therefore cannot directly measure Milvus and Zilliz Cloud batch-search workloads with NQ greater than 1.

I added `nq` to the common case configuration and CLI (`--nq`, or `nq` in YAML), with the existing behavior preserved by a default of 1. Milvus advertises batch-search support through a client capability and implements one SDK search call for each fixed-size query batch; Zilliz Cloud inherits this implementation. The concurrent runner wraps through the query dataset to keep batches full, counts successful query vectors in `qps` and `conc_qps_list`, records latency per successful batch request, and excludes failed or incomplete batches from successful query-vector counts. Serial latency and recall remain single-query measurements, while unsupported backends and workloads reject `nq > 1` before data loading.

A CLI unit test verifies that the configured NQ reaches the generated task configuration. Live Milvus validation used a custom runner's `run_by_dur()` path with NQ=64 on Cohere and BioASQ at 1M/10M and topK=10/100; all eight combinations reported no failed batches. This was not an end-to-end run through the standard CLI, and Zilliz Cloud was not live-tested.


yanbinyang:
/assign @xuanyang-cn

Hi, when you have time, could you please take a look? The workflow is awaiting maintainer approval. Thanks!

yanbinyang:
Hi @jamesgao-jpg, I have rebased this PR on main and added regression coverage in 40df7e3. All CI checks pass. When convenient, could you please take a look? Thanks!

PR #876, opened 2026-09-16
feat: add search level to Milvus AutoIndex

### LINKED ISSUE #875: [Feature Request] Allow an optional search level in milvusautoindex
Milvus already exposes an AutoIndex search `level` through its
[Go SDK](https://github.com/milvus-io/milvus/blob/master/client/index/auto.go).
VectorDBBench's Milvus `AutoIndexConfig` and `milvusautoindex` CLI do not expose
this parameter, so it cannot be set when benchmarking servers that implement it.

Add `--level` (YAML: `level`) from 1 to 10, forwarding it as `search_params.params.level`.
If omitted, leave search parameters unchanged. Index creation should not change.
The server determines how the level is used; this only adds client-side parameter support.

## Example

Search an existing, loaded collection containing the selected dataset and an
AUTOINDEX index:

```shell
vectordbbench milvusautoindex \
  --uri http://localhost:19530 --collection-name VDBBench \
  --case-type Performance768D1M --level 2 --k 10 \
  --skip-drop-old --skip-load --skip-search-concurrent
```

yanbinyang:
Fixes #875. Milvus AutoIndex supports an optional search level on compatible deployments, but VectorDBBench did not expose it through its configuration or CLI. As a result, CLI and YAML runs could not reproduce or record level-specific AutoIndex searches.

I added an optional `level` field to `AutoIndexConfig` and exposed it through `milvusautoindex` as `--level`, or `level` in YAML, with validation from 1 through 10. The search parameter includes the level only when it is configured, so omitting the option preserves the server default and index creation remains unchanged. The FLAT command now uses a separate option type so `milvusflat` does not accept an AutoIndex-only parameter.

Added a CLI unit test that verifies the configured level reaches the Milvus AutoIndex case configuration. A live search-only run on a level-capable Milvus deployment also reproduced the previous recall across all eight Cohere and BioASQ cases at 1M and 10M, each with K=10 and K=100.


yanbinyang:
/assign @xuanyang-cn

Hi, when you have time, could you please take a look? The workflow is awaiting maintainer approval. Thanks!

vectordb_bench/backend/clients/milvus/config.py
```diff
@@ -53,6 +53,7 @@ def parse_metric(self) -> str:
 
 class AutoIndexConfig(MilvusIndexConfig, DBCaseConfig):
     index: IndexType = IndexType.AUTOINDEX
+    level: int | None = Field(default=None, ge=1, le=10)
```
jamesgao-jpg: Parity suggestion: Zilliz Cloud already has its own `AutoIndexConfig` with a separate `level: int = 1` (`zilliz_cloud/config.py:30`), and `zillizautoindex --level` is `type=str` with no range check. Consider making the Zilliz class a thin subclass of this one (`class AutoIndexConfig(MilvusAutoIndexConfig)`) so both providers share one `level` field + the 1–10 validation, keeping Zilliz's `level=1` default and always-send `search_param()` so existing `zillizautoindex` runs/results are unchanged. Boundary: `ZillizCloudFtsConfig` must stay on `MilvusFtsConfig` (BM25/sparse machinery) with its own `level: int = 1` — it must not inherit this dense config.

vectordb_bench/backend/clients/milvus/cli.py
```diff
@@ -100,7 +100,16 @@ class MilvusTypedDict(TypedDict):
     ]
 
 
-class MilvusAutoIndexTypedDict(CommonTypedDict, MilvusTypedDict): ...
+class MilvusAutoIndexTypedDict(CommonTypedDict, MilvusTypedDict):
+    level: Annotated[
```
jamesgao-jpg: Consider extracting this option into a shared TypedDict fragment (same pattern as `HNSWFlavor3` / `IVFFlatTypedDictN` in `vectordb_bench/cli/cli.py`) so `zillizautoindex` reuses the same `--level` definition. Zilliz's current `--level` is `type=str` with a manual `int()` conversion — non-numeric input crashes with a traceback and out-of-range values pass through unvalidated — while `click.IntRange(1, 10)` rejects both cleanly.

jamesgao-jpg (commented):
Thanks for the contribution!

I think the idea is good, but if you have some free time, it will be really great if we can fix zilliz's autoindex together with this pr. 

1. Unify the search-level entry point with Zilliz Cloud: make `zilliz_cloud/config.py`'s `AutoIndexConfig` a thin subclass of the renewed Milvus `AutoIndexConfig` so both providers share one `level` field and its 1–10 validation, while keeping Zilliz's `level=1` default / always-send behavior so existing `zillizautoindex` runs and result files are unchanged. `ZillizCloudFtsConfig` stays on `MilvusFtsConfig` with its own `level: int = 1`. This warrants a refactor but I will do it later.

2. Share the CLI option as a TypedDict fragment so `zillizautoindex --level` gets the same `IntRange(1,10)` validation (it is currently `type=str` with a manual `int()` conversion).

3. Add regression coverage for the omitted-level and `milvusflat`-rejects-`--level` contracts, plus out-of-range rejection.

4. README placement nit + note that `zillizautoindex` shares `--level`.

Thanks in advance!

tests/test_milvus_zilliz_cli.py
```diff
@@ -55,6 +55,31 @@ def fake_run(**kwargs):
...
+            "--uri",
+            "http://localhost:19530",
+            "--level",
+            "2",
+            "--dry-run",
+        ],
+    )
+
+    assert result.exit_code == 0, result.output
+    assert captured["db_case_config"].level == 2
```
jamesgao-jpg: Set-path coverage only. The PR's two backward-compat claims are untested: (1) omitted `--level` → `level is None` and `search_param()` has no `params` key (server default preserved); (2) `milvusflat` rejects `--level` (the stated reason for the `MilvusFlatTypedDict` split). Suggest sibling assertions for the omitted case, the `MilvusFlat --level` rejection, and out-of-range `--level 0` / `11`.

README.md
```diff
@@ -133,6 +133,10 @@ vectordbbench zillizautoindex \
   <other options>
 ```
 
+`milvusautoindex` accepts `--level` from 1 to 10 (YAML: `level`) on servers that
```
jamesgao-jpg: Minor: this paragraph sits inside the `--note-file` discussion right after a `zillizautoindex` example. Suggest moving it next to a `milvusautoindex` example, and (if the Zilliz subclass suggestion lands) noting that `zillizautoindex` shares the same `--level` (1–10).

vectordb_bench/backend/clients/milvus/config.py
yanbinyang: Addressed in aac8019: Zilliz Cloud AutoIndexConfig now subclasses the Milvus AutoIndex config and shares the same constrained level type. Its default remains 1 and the inherited search_param still always sends it. ZillizCloudFtsConfig remains on MilvusFtsConfig.

vectordb_bench/backend/clients/milvus/cli.py
yanbinyang: Addressed in aac8019: extracted a shared AutoIndexLevelTypedDict with click.IntRange(1, 10). Both milvusautoindex and zillizautoindex reuse it, and the Zilliz manual string-to-int conversion has been removed.

README.md
yanbinyang: Addressed in aac8019: moved the level documentation next to a milvusautoindex example and documented the shared 1-10 range plus the different omitted-value behavior for Milvus and Zilliz Cloud.

tests/test_milvus_zilliz_cli.py
yanbinyang: Addressed in aac8019: added focused coverage for omitted Milvus level/no params, milvusflat rejecting --level, and Milvus rejecting out-of-range values 0 and 11.

yanbinyang:
Hi @jamesgao-jpg , I’ve addressed all review comments in aac8019 and replied inline. When convenient, could you please take another look?

PR #879, opened 2026-09-21
Read turbopuffer query result ids from the raw response

owenhalpert:
The turbopuffer client builds a pydantic model per returned row, but only the ids are used. Interleaved A/B on the same queries, the model construction is about 28% of client CPU per query at top_k=100 (1.30ms -> 0.94ms), rising to 47% at top_k=1000.

Reading the ids off the raw response instead: 2,910 -> 3,700 QPS on Cohere 1M (768d), recall unchanged at 0.9045. No new dependencies, `loads` is already imported.

Marking as draft while I confirm there are no edge cases with the multitenant and payload profile paths.


owenhalpert:
Closing this — my benchmark numbers were not measured rigorously enough. Re-running as a proper alternating A/B shows a much smaller difference than the PR claimed. Will revisit if it holds up.

